PR #21, opened 2015-07-13
Add plugin support

grooverdan:
Adds MySQLServerCapabilities.PLUGIN_AUTH to the default capabilities of
the client.

Correctly decodes server announced capabilities by reading the second
short value representing the high values.

Parses out the plugin name from the server protocol.

Correctly produces a MySQLClientAuthPacket based on server capabilities
and extends this to potentially allow for other client authentication
plugins in future.


rusher:
Thanks, i will test that tomorrow when adding test cases

rusher:
closing this request, because your second request https://github.com/MariaDB/mariadb-connector-j/pull/22 is more complete (and merged!)

PR #22, opened 2015-07-16
CONJ 164 plugin implementation

grooverdan:
Added plugin implementation for mysql_clear_password and mysql_native_password. dialog partially there.


rusher:
thanks for mergin will failover version, there was a lot to do !

grooverdan:
after checking dialog plugin needs a bit more work. I'll try to add test cases too.

rusher:
Seem's good ! 

Maybe the "processPlugin" method can be put in a new Object like MySQLGreetingReadPacket ? MySQLProtocol is big enough :)

So, i'll wait for change on dialog and test cases, or maybe you can send a new request without the dialog part as first Request, i will add test cases  ?

grooverdan:
Added the bits to plugin dialog to correct it, but still needs a callback

> Maybe the "processPlugin" method can be put in a new Object like MySQLGreetingReadPacket ? MySQLProtocol is big enough :)

Yes. I imagine once https://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html  (defaultAuthenticationPlugin) is implemented a consistent form for the rest of the plugins would be a obvious.

> or maybe you can send a new request without the dialog part as first Request,

Is what's here now acceptable?

> i will add test cases ?

Sounds good. Here is what I did for a python implementation also extending  the client implementation to support plugins which does include tests. PyMySQL/PyMySQL/pull/353
The only thing I haven't done is the following which would need a dialog callback implemented.
<pre>
INSTALL PLUGIN two_questions SONAME 'dialog_examples';
INSTALL PLUGIN three_questions SONAME 'dialog_examples';
</pre>

rusher:
Thanks for you contribution daniel.

diego.

grooverdan:
take it under the BSD-new license

rusher:
Close since this part has been totally rewritten, and functionnality implemented

PR #25, opened 2015-08-02
getProtocol() scope, LOCAL INFILE option, SSL fix, and SSL Cert Enhancement

sehrope:
This PR is split out into four separate commits:

1. Change `MySQLConnection.getProtocol()` method scope to "public". 
    - This allows using it to issue a query cancellation.
2. Add a new option `allowLocalInfile` defaulting to "true"
    - When disabled (*i.e. set to `false`*) it throws an error if a `... LOCAL INFILE ...` query is executed without first specifying an InputStream for it. This prevents a query from specifying an arbitrary file path via a URL syntax.
    - The default is set to "true" so that the existing behavior is preserved (i.e. no compatibility issue).
3. Change SSLSocketFactory to set `autoClose` to true when creating the SSLSocket
    - This fixes SSL connections from hanging when they're closed.
    - See this [JDK bug](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6932592) for more info
4. Enhance custom SSL certificate handling to allow specifying multiple certificates.
    - The existing SSL certificate code (*originally written by me :)*) only handles loading a single certificate. This patch allows more than one certificate to be specified.
    - A use case for this is having both the old and new versions of the CA cert for your database. That way when the SSL certificate on your server is rotated there won't be any downtime for your app.

Let me know if I should split them out into separate PRs (rather than just separate commits). I figured this would be easier as it avoids creating a bunch of intermediate merge commits.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/common/DefaultOptions.java
+++ b/src/main/java/org/mariadb/jdbc/internal/common/DefaultOptions.java
@@ -118,6 +118,11 @@ public enum DefaultOptions {
     USE_OLD_ALIAS_METADATA_BEHAVIOR("useOldAliasMetadataBehavior", Boolean.FALSE, "1.1.9"),
 
     /**
+     *  If set to 'false', exception thrown during LOCAL INFILE if no InputStream has already been set.
+     */
+    ALLOW_LOCAL_INFILE("allowLocalInfile", Boolean.TRUE, "1.2.1"),
+
+    /**
      * var=value pairs separated by comma, mysql session variables, set upon establishing successful connection
      */
     SESSION_VARIABLES("sessionVariables", "1.1.0"),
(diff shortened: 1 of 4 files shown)


sehrope:
FYI about half of the Travis-CI test matrix failed but it's not related to this PR. It's some dependency issue fetching/installing the server to test against:

>    W: Failed to fetch http://ftp.igh.cnrs.fr/pub/mariadb/repo/10.0/ubuntu/dists/precise/main/binary-i386/Packages  500  Internal Server Error

The tests all pass for the environments that get past the server installation and actually run the tests.

sehrope:
I rebased the PR atop the Travis fixes (PR #26). Now all the tests pass.

rusher:
I don't like the idea of changing the protocol access to public. 
1 - This object is internal and changed a lot. If some rely on it , it may break code.
2 - The query cancellation is standard in JDBC using Statement.cancel(), better to keep it standard too.  

Good to have add allowLocalInfile option

> (originally written by me :))

SSL original writer ! good, keep the good work so !

sehrope:
> I don't like the idea of changing the protocol access to public. 

That's fair. I've removed that commit from the PR. I'll figure out a workaround as accessing the Protocol directly is convenient.

> SSL original writer ! good, keep the good work so !

Well not the whole SSL stack, just the piece that validates SSL certificates. Thanks though! :)

sehrope:
Awesome. Thanks for merging!

PR #28, opened 2015-08-06
Clean up unused imports, whitespace, and newlines

sehrope:
Just a bunch of misc clean up of whitespace, trailing newlines, and unused imports.


sehrope:
I rebased this atop the latest. Would be nice to get it applied.

rusher:
done now in code in all files (156) with constructor on top of file and so on.
Thanks for pointing that out !

Issue #32, opened 2015-08-09
SSL Tests Not Run on Travis-CI


sehrope:
The SSL tests are not being run by the automated testing on Travis-CI. The install script does not configure the servers for SSL and the client JUnit tests silently skips the tests if SSL is not enabled on the server.

rusher:
close after merge of your pull request https://github.com/MariaDB/mariadb-connector-j/pull/33

PR #33, opened 2015-08-09
Fix Travis-CI SSL

sehrope:
This patch fixes the Travis-CI server setup to configure SSL for the server versions that are tested. It also fixes some of the SSL tests so they run successfully (*the null password was breaking them*).

It's layered atop PR #31 (*as without that none of the trustServerCertificate tests would work*) and closes out #32.

### DIFF
--- a/.travis/before_test.sh
+++ b/.travis/before_test.sh
@@ -50,6 +50,18 @@ max_allowed_packet=$MAX_ALLOWED_PACKET
 innodb_log_file_size=$INNODB_LOG_FILE_SIZE
 END
 
+# Generate SSL files:
+sudo .travis/gen-ssl.sh mariadb.example.com /etc/mysql
+sudo chown mysql:mysql /etc/mysql/server.crt /etc/mysql/server.key /etc/mysql/ca.crt
+
+# Enable SSL:
+sudo tee /etc/mysql/conf.d/ssl.cnf << END
+[mysqld]
+ssl-ca=/etc/mysql/ca.crt
+ssl-cert=/etc/mysql/server.crt
+ssl-key=/etc/mysql/server.key
+END
+
 sudo mysql -u root -e "SET GLOBAL innodb_fast_shutdown = 1"
 sudo service mysql stop
 sudo rm -f /var/lib/mysql/ib_logfile*
--- a/.travis/gen-ssl.sh
+++ b/.travis/gen-ssl.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+set -e
+
+log () {
+  echo "$@" 1>&2
+}
+
+print_error () {
+  echo "$@" 1>&2
+  exit 1
+}
+
+print_usage () {
+  print_error "Usage: gen-ssl-cert-key <fqdn> <output-dir>"
+}
+
+gen_cert_subject () {
+  local fqdn="$1"
+  [[ "${fqdn}" != "" ]] || print_error "FQDN cannot be blank"
+  echo "/C=/ST=/O=/localityName=/commonName=${fqdn}/organizationalUnitName=/emailAddress=/"
+}
+
+main () {
+  local fqdn="$1"
+  local sslDir="$2"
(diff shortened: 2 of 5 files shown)


rusher:
Very good job sehrope.

There is something to digg before validating your code : there seems to be a random error when using mysql database with error "javax.net.ssl.SSLException: Unsupported record version Unknown-0.0".
I will see to that now.

rusher:
not seen the error for the moment. 
it will have to wait for a few days.

sehrope:
Yes I think it may be a transient issue with the Travis-CI environment. I noticed that the MySQL/MariaDB installs what fail from time to time as well as intermittent errors connecting some times as well. 

I just kicked off another run on Travis to see if it's reproducible.

sehrope:
I'm seeing two types of errors on Travis running the SSL tests. 

### `Unsupported record version Unknown-0.0`
This is the one you ran into as well. I saw it appear once for two different JDK versions but not always for those versions. Re-running the failed jobs has them eventually succeed. Here's the latest Travis build: https://travis-ci.org/sehrope/mariadb-connector-j/jobs/74819079

I'm not sure the exact cause but it seems like a race condition. Maybe the server isn't ready by the time the tests start connecting to it.

### oraclejdk7 JCE error
Here's the second error:

    java.sql.SQLNonTransientConnectionException: Could not connect to localhost:3306: java.lang.RuntimeException: Could not generate DH keypair
        at org.mariadb.jdbc.internal.SQLExceptionMapper.get(SQLExceptionMapper.java:136)
        [... truncated ...]
    Caused by: org.mariadb.jdbc.internal.common.QueryException: Could not connect to localhost:3306: java.lang.RuntimeException: Could not generate DH keypair
        at org.mariadb.jdbc.internal.mysql.MySQLProtocol.connect(MySQLProtocol.java:424)
        [... truncated ...]
    Caused by: javax.net.ssl.SSLException: java.lang.RuntimeException: Could not generate DH keypair
        at sun.security.ssl.Alerts.getSSLException(Alerts.java:208)
        [... truncated ...]
    Caused by: java.lang.RuntimeException: Could not generate DH keypair
        at sun.security.ssl.DHCrypt.<init>(DHCrypt.java:136)
        [... truncated ...]
    Caused by: java.security.InvalidAlgorithmParameterException: Prime size must be multiple of 64, and can only range from 512 to 1024 (inclusive)
        at com.sun.crypto.provider.DHKeyPairGenerator.initialize(DHKeyPairGenerator.java:120)
        at java.security.KeyPairGenerator$Delegate.initialize(KeyPairGenerator.java:658)
        at sun.security.ssl.DHCrypt.<init>(DHCrypt.java:127)
        ... 51 more

This is a JDK issue. Out of the box Oracle JDK7 doesn't allow for large key sizes. The unlimited JCE policy needs to be installed. More info hiere: https://github.com/travis-ci/travis-ci/issues/832

We should be able to add this to the pre-install script (i.e. detect Oracle JDK and download/install it). The Travis-CI cookbook does exactly that here: https://github.com/travis-ci/travis-cookbooks/blob/ff28e574541d34669f90c2bd9f21a297307f31f3/ci_environment/java/recipes/oraclejdk7.rb#L45

sehrope:
Anyway I think this should get merged in. The tests are  better off than they were before and the remaining issue with oraclejdk7 can be handled separately.

Probably should have a release soon too as the most recent release (v1.2.0) was shipped with the broken trustServerCertificate SSL support and this fixes it.

rusher:
I'm testing this now with some addon :
- I've added JCE to travis, and that correct the "Could not generate DH keypair" problem.

- try now to resolve the "Unsupported record version Unknown-0.0" problem. This append when connecting to a MYSQL DB, never with a MariaDB DB . It seems related to compilation of MYSQL with YaSSL (community version). 

A correction released will be delivered during the week that will wait for this issue.

rusher:
when using MySQL DB, you where right about the race condition, adding a "sleep 5" to travis when using MySQL DB correct the problem., so, merging now

PR #54, opened 2015-12-02
Fix unit of default value of retriesAllDown in the documentation

ksoichiro:
As described in "Description" of "Failover / high availability options" section,
"retriesAllDown" means not seconds but the maximum number of connection attempts,
so I removed "seconds" from it.


rusher:
indeed !

PR #56, opened 2015-12-11
Fix blacklist race condition in AbstractMastersListener and sensitivity to system clock changes

jentfoo:
I have been discussing these issues and others with @rusher via email.  These are two issues which I have fixed in the fullcontact/mariadb-connector-j fork that do not depend on my threadly library, and thus are trivial to merge now.

I figured I would send these first because it's both trivial to separate out, and so I don't overload you guys with the next fix.

Check the commit messages for more details to each issue, and how they were fixed.

### DIFF
--- a/src/test/java/org/mariadb/jdbc/failover/AuroraFailoverTest.java
+++ b/src/test/java/org/mariadb/jdbc/failover/AuroraFailoverTest.java
@@ -7,6 +7,7 @@ import org.mariadb.jdbc.internal.util.constant.HaMode;
 import java.sql.*;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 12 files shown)


src/test/java/org/mariadb/jdbc/failover/AuroraFailoverTest.java
```diff
@@ -119,21 +120,21 @@ public void pingReconnectAfterRestart() throws Throwable {
...
 
-            long stoppedTime = System.currentTimeMillis();
             try {
                 st.execute("SELECT 1");
             } catch (SQLException e) {
                 //normal exception
             }
             restartProxy(masterServerId);
-            long restartTime = System.currentTimeMillis();
+            long restartTime = System.nanoTime();
```
jentfoo: Even though we are not concerned about time changes in unit tests, nano time must be used since System.currentTimeMillis() may varry slightly from internal nano time usage.  This seems more accurate than just making the time checks have a larger verification window.

FWIW, these tests were not failing for me before the change.  I just know that from experience failing to change this may cause random unit test failures.

src/test/java/org/mariadb/jdbc/failover/AuroraFailoverTest.java
```diff
@@ -119,21 +120,21 @@ public void pingReconnectAfterRestart() throws Throwable {
...
 
             boolean loop = true;
             while (loop) {
                 if (!connection.isClosed()) {
                     loop = false;
                 }
-                if (System.currentTimeMillis() - restartTime > 15 * 1000) {
+                long duration = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - restartTime);
```
zhanhb: maybe TimeUnit.SECONDS is better.

jentfoo: I had considered that, let me know what your preference is.  You will see in the main code I did convert most to seconds.

rusher:
Nice job @jentfoo !

This PR will be a the next correction version 1.3.4.

So please indicate in a comment below that you are contributing the code of the whole pull request under the BSD-new license (the other alternative is to sign the code contribution agreement which can be found here: https://mariadb.com/kb/en/mariadb/mca/).

jentfoo:
This entire pull request contribution is under the BSD-new license.

PR #57, opened 2015-12-15
MastersListener: Fixed several issues around thread pool usage

jentfoo:
A bit of a sizable commit, several things are included, though they are all closely related.  Because this one may have less obvious changes than my previous pull request, please ask questions.

This pull request will fix all the defects I found and fixed in our fork (https://github.com/fullcontact/mariadb-connector-j).  There is one sizable difference between these.  I tried to get the thread count reduction also to work well, but it ended up being too complicated with java.util.concurrent.ScheduledThreadPoolExecutor.  So this has the same thread #'s that current master has.  However as you will see in one of my bullets and the changes, I did add a way for library users to change how the pool's are constructed and live.  This should allow us to override this behavior within the library so that we can at least get better thread counts (or anyone else who wants to include threadly).  Let me know if there are any questions or concerns, but we would at least like to be able to use a substantially better pool than what java.util.concurrent provides.

The following defects in AbstractMastersListener were fixed:
* Thread leakage which results in threads to continue to grow without bounds till the server runs out of memory
    The issue here is that ScheduledThreadPoolExecutor does not cleanup started threads once allowed to be garbage collected (arguably a defect in the java classpath, this is due to the inner class of ThreadPoolExecutor "Worker" not being static, and thus holding on to a reference of the pool).

    The fix here was to use a common pool for the ping and fail loop that has a normal start/shutdown lifecycle.

* Fixed issue where FailLoop may fail to stop, as well as allowing multiple FailLoop's to be started
    Condition possible when:
    Thread A enters 'launchFailLoopIfNotlaunched' and progresses to the point where it has transitioned the atomic boolean, but then one of two conditions happens:
         Either thread A is de-scheduled by the OS before 'scheduledFailover' is set
         ...OR... 'scheduledFailover' is set, but since it's not volatile the update is not witnessed by thread B (sadly because of the above condition this could not be solved by just setting this as volatile)
    Thread B then enters 'stopFailover', it transitions the AtomicBoolean back to false, but does not see the set 'scheduledFailover', so it does not remove the task.  I suspect this is why your doing the null check, previously when you would have got a NullPointerException, now your just leaking the task.

    This problem would be further compounded now that it is possible for another invocation to schedule a second task if 'launchFailLoopIfNotlaunched' is invoked.  The null check that existed here likely was a witness of this race condition.

    The solution here was to use an "AtomicReference" hold an instance of the "FailLoop", which can be set before scheduled.  Once set there, we let the FailLoop track its own scheduled future.  It may need to block to wait for the future to be set in a very tight condition, but otherwise unscheduling is handled interior to the task.

 Other changes:
 * Also changed was the ability to allow library users to set their own thread pools.  java.util.concurrent's thread pools are deficient in many ways (performance being one of them).  This will allow us (and others if desired) to provide their own implementations.
 * Threadly was added as a TEST dependency.  I did this because when writing unit tests for the new 'SchedulerServiceProviderHolder' I was being lazy and did not want to re-implement "TestRunnable"'s trivial blocking/verification.  I also updated a couple unit tests, one that was purely simplified, but another where I showed the use of threadly's "TestableScheduler" for deterministic execution.  If there are any concerns here, ideally I would like to just remove the test which made the dependency worth it (it's testing pretty trivial code anyways).

As with before, this is licensed under the BSD-new license.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
+++ b/src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
@@ -85,7 +85,7 @@ public abstract class AbstractMastersListener implements Listener {
     public final UrlParser urlParser;
     protected AtomicInteger currentConnectionAttempts = new AtomicInteger();
     protected AtomicBoolean currentReadOnlyAsked = new AtomicBoolean();
-    protected AtomicReference<FailLoop> runningFailLoop = new AtomicReference<>();
+    private AtomicReference<FailLoop> runningFailLoop = new AtomicReference<>();
     protected Protocol currentProtocol = null;
     protected FailoverProxy proxy;
     protected long lastRetry = 0;
@@ -432,6 +432,7 @@ public abstract class AbstractMastersListener implements Listener {
                 } else {
                     scheduledFuture.cancel(false);
                     scheduledFuture = null;
+                    return;
                 }
             }
         }
--- a/src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
+++ b/src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
@@ -51,6 +51,8 @@ OF SUCH DAMAGE.
 
 import org.mariadb.jdbc.HostAddress;
 import org.mariadb.jdbc.UrlParser;
+import org.mariadb.jdbc.internal.util.SchedulerServiceProviderHolder;
+import org.mariadb.jdbc.internal.util.SchedulerServiceProviderHolder.SchedulerProvider;
 import org.mariadb.jdbc.internal.util.dao.QueryException;
 import org.mariadb.jdbc.internal.query.MariaDbQuery;
 import org.mariadb.jdbc.internal.query.Query;
@@ -64,12 +66,13 @@ import java.sql.SQLException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
 
 
 public abstract class AbstractMastersListener implements Listener {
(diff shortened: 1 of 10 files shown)


src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
```diff
@@ -96,6 +99,24 @@ protected AbstractMastersListener(UrlParser urlParser) {
...
+    protected void shutdownScheduler() {
+        schedulerProvider.shutdownScheduler(scheduler);
+        FailLoop runningLoop = stopFailover();
+        if (runningLoop != null) {
+            runningLoop.blockTillTerminated();
+        }
+    }
+    
+    @Override
+    protected void finalize() throws Throwable {
```
jentfoo: I forgot to mention I modified the checkstyle to allow this finalizer.  This seems like a pretty important check to me, since it could result in leaked threads if not invoked (leaked threads were also possible before if the scheduler was not shutdown, but not being verified)

rusher:
Hello jentfoo, 
I like your code, it's elegant.
I'm doing the full test tour of failover implementation with this PR, and add various failover test. It will take some time, but hoping to make a v 1.3.4 soon with this.

src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
```diff
@@ -352,7 +416,28 @@ public FailLoop(Listener listener) {
...
+        }
+
+        // while scheduledFuture may be set from another thread
+        // it is expected only one thread will unschedule
+        @Override
+        public void unscheduleTask() {
+            while (true) {
+                if (scheduledFuture == null) {
+                    // not scheduled yet, should be a rare condition, just loop
+                    Thread.yield();
```
rusher: If failLoop reconnect the failed connection, the stopFailover() inside the loop will loop inside this while, and never freed the thread.

src/main/java/org/mariadb/jdbc/internal/failover/AbstractMastersListener.java
```diff
@@ -352,7 +416,28 @@ public FailLoop(Listener listener) {
...
+                } else {
+                    scheduledFuture.cancel(false);
+                    scheduledFuture = null;
```
rusher: a "return" after this seems good : If failLoop reconnect the failed connection, the stopFailover() (actual line 451) will loop inside this while, and never freed the thread.

jentfoo: I think I see the issue, though a return statement is also necessary, there seems to be an additional issue here.

The issue is that this was expected to only be invoked from stopFailover().  The reason this is necessary is because we assume at this point that a future _should_ exist, or will exist shortly.  stopFailover() is our thread safety to ensure that this assumption is accurate.

The issue is that this could also be invoked by "blockTillTerminated()".  A couple obvious solutions come to mind:
* we could make unscheduleTask defer to stopFailover(), and have stopFailover() invoke a new function with this same logic).
* We could move the AtomicReference into an AtomicBoolean inside this class, and have a final FailLoop instance that we defer to for scheduling.  This would require us to change our termination behavior to allow a task to be reset to a runnable state.

Let me know if you have a preference, the first option would have the less code change and would fit rather nicely.  The second option I think would be ultimately a bit cleaner, as it would reduce the dependence on the parent class.

For now I will probably just update the PR with the first option since it should be quick and easy to do, but let me know if you have a preference.

jentfoo: I just pushed a fix which includes the return.  I also just made the AtomicReference private, since it seems like we are never invoking "blockTillTerminated" outside of stopFailover anyways.  So we are already guarding that behavior and wont have the above condition.  We just need to be sure that attempts to stop the fail loop always go through "stopFailloop", and not to the loop directly.

I will say, it's a bit confusing (I was confused when I made the above comment), it probably would be worth doing my second bullet at some point to move more of the FailLoop thread safety into FailLoop, and just have a single reference which we invoke to schedule and unschedule itself.  Just a thought.

Thanks for the feedback!  what a silly mistake on my part.  Let me know if you have any other issues or questions.

jentfoo:
Closing, this has been replaced by #60

PR #59, opened 2015-12-17
Fixes to be able to retrieve generated keys from execute method

tvburger:
Is there a reason to have the generic form not able to return the generated keys?
Some applications are using this driver, but they have abstracted beyond the executeUpdate and executeQuery, and thus can't use the executeUpdate.

This issue was introduced since 1.3.X but not existent in 1.2.X. Any reasons why?


rusher:
JDBC permit to retrieve previous insert id's by the Statement.getGeneratedKeys() command.
By default id's are not retrieved. When executing query, additional parameters Statement.NO_GENERATED_KEYS or Statement.RETURN_GENERATED_KEYS indicate if id's must be retrieved, like in command Statement.executeUpdate(String sql, int autoGeneratedKeys).
That permit to save memory. 

If needed, an option "alwaysAutoGeneratedKeys" permit to retrieved generatedKeys without setting those informations. (if set to true, insert id's will be retrieved even if no autoGeneratedKeys parameter or Statement.NO_GENERATED_KEYS is set.)

PR #63, opened 2016-01-07
SchedulerServiceProvider: Return interface rather than implementation

jentfoo:
I must have missed that in the fixed size case we were returning the implementation, rather than the interface.
Changing to the interface allows for easier extension and other scheduler implementations here.


rusher:
good point.

MIke, are you ok to submit your code under the BSD-new license ?

jentfoo:
whoops, forgot to mention that.  Yes, this is all submitted under the BSD-new license

PR #64, opened 2016-01-07
MasterSlavesListener: Fix for race condition in switchReadOnlyConnection

jentfoo:
An AtomicBoolean does not provide a sufficent enough concurrency barrier here.
Instead I switched it to a volatile boolean + an Object that we synchronize on when the boolean is to be updated.

I am pretty sure I hit this today.  We are starting to use our aurora read replica more and more, so I am not unsurprised I might have uncovered this (or possibly another) race condition.  I had a condition where a connection, fresh from the pool (which I know had the read only state set back after it was done), complained as it tried to hit the read replica.

I hunted around for what might be the cause, and so far this is the only thing I have found.  If you have other ideas where this issue might be, please point me to where to look.

I will comment at the line where I believe the race condition existed.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/failover/impl/MastersSlavesListener.java
+++ b/src/main/java/org/mariadb/jdbc/internal/failover/impl/MastersSlavesListener.java
@@ -207,7 +207,7 @@ public class MastersSlavesListener extends AbstractMastersSlavesListener {
         checkWaitingConnection();
         //if connection is closed or failed on slave
         if (this.currentProtocol != null
-                && (this.currentProtocol.isClosed() || (!currentReadOnlyAsked.get() && !currentProtocol.isMasterConnection()))) {
+                && (this.currentProtocol.isClosed() || (!currentReadOnlyAsked && !currentProtocol.isMasterConnection()))) {
             preAutoReconnect();
         }
     }
@@ -346,7 +346,7 @@ public class MastersSlavesListener extends AbstractMastersSlavesListener {
             masterProtocol.close();
         }
 
-        if (!currentReadOnlyAsked.get() || isSecondaryHostFail()) {
+        if (!currentReadOnlyAsked || isSecondaryHostFail()) {
             //actually on a secondary read-only because master was unknown.
             //So select master as currentConnection
             try {
@@ -408,7 +408,7 @@ public class MastersSlavesListener extends AbstractMastersSlavesListener {
         }
 
         //if asked to be on read only connection, switching to this new connection
-        if (currentReadOnlyAsked.get() || (urlParser.getOptions().failOnReadOnly && !currentReadOnlyAsked.get() && isMasterHostFail())) {
+        if (currentReadOnlyAsked || (urlParser.getOptions().failOnReadOnly && !currentReadOnlyAsked && isMasterHostFail())) {
             try {
                 syncConnection(currentProtocol, newSecondaryProtocol);
             } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/mariadb/jdbc/internal/failover/impl/MastersSlavesListener.java
```diff
@@ -434,80 +434,88 @@ public void lockAndSwitchSecondary(Protocol newSecondaryProtocol) throws QueryEx
      */
     @Override
     public void switchReadOnlyConnection(Boolean mustBeReadOnly) throws QueryException {
-        if (mustBeReadOnly != currentReadOnlyAsked.get() && currentProtocol.inTransaction()) {
+        if (mustBeReadOnly != currentReadOnlyAsked && currentProtocol.inTransaction()) {
             throw new QueryException("Trying to set to read-only mode during a transaction");
         }
         checkWaitingConnection();
-        if (currentReadOnlyAsked.compareAndSet(!mustBeReadOnly, mustBeReadOnly)) {
```
jentfoo: Here is where I believe to be the problem...

While the compareAndSet can ensure that the currentReadOnlyAsked is in a consistent state, it does not guard us through the logic contained within this if condition.

Consider if you will, two threads A & B (I am assuming two threads can hit this in parallel?):

Thread A switches the state (for sake of example false -> true)...CAS works fine
Thread A checks currentReadOnlyAsked and gets true, starts doing work to switch to secondary
Thread A is now forced to yield by OS scheduler, or possibly is blocked on proxy.lock (probably more likely)
Thread B switches the state(true -> false)...CAS again works fine, even though Thread A has not finished changing state
Thread B is able to complete update without blocking, very possible since proxy.lock is not a fair lock (and I am NOT recommending it should be)
Thread A now wakes up and switches connection back to read only

I have solved this by creating a memory barrier for this entire section.  However I am not 100% confident that there is not potential for a deadlock around proxy.lock (I don't see one, but proxy.lock is a nasty beast).  I did email @rusher to discuss the complexity of proxy.lock/MariaDbConnection.lock (same lock), so maybe we want to fix this in a different way with more intense locking changes.

Thoughts?

basilmusa: @jentfoo I'm not sure why you switched from an AtomicBoolean to a volatile and a synchronized object combination. If you are sure that AtomicBoolean is defunct, I would say remove it. But I don't think this is the case. I would recommend leaving the AtomicBoolean as is, until the problem is further investigated.

Further reads on AtomicBoolean versus volatile variables in java: 
http://stackoverflow.com/a/3787435/367988
http://stackoverflow.com/a/3787093/367988

rusher:
> I had a condition where a connection, fresh from the pool (which I know had the read only state set back after it was done), complained as it tried to hit the read replica.
Technically using the same connection in multiple thread must work, but that is not the good way to use connection : the mysql protocol is synchronous in a sense that the same connection does not execute 2 queries in parallel.  thus there is no real advantage of using the same connection from multiple threads.
I assume that if you use aurora, you are using pooling. You can retrieve a new connection on the other thread, and you will never have this kind of problem. (And performance will be a lot better !)

but multithread must work too, and concerning a possible race condition : you are right, there can be a race condition here.
but use 2 locks (a synchronized + lock just after) may not be the best way to do this, since there will be overhead for the "normal" mono thread use case.

src/main/java/org/mariadb/jdbc/internal/failover/impl/MastersSlavesListener.java
jentfoo: @basilmusa The reason for using a volatile is to avoid the overhead of synchronization during the read operations.  If we were not using volatile, even if we used a synchronized block to guard the writes, we would need to synchronize for the reads.  If we did not synchronize on those reads we may see an old version (ie a version on the local CPU cache if no context switches occurred).  It's unlikely, but that is how the java memory model is defined.

If you want references for this, I can dig them up.  But that's was the reason.

jentfoo: Oh, if your suggesting just keep the AtomicBoolean...yea, My personal option (style), is don't use AtomicX unless you need compare and swap operations, or unless you can take advantages of Atomic's weak set's.  Otherwise it's not worth the (granted very minor) memory usage.  It also may cause confusion by making it seem like a CAS operation may be safe to do on that variable, when instead it must be guarded by the lock.

So more style, than necessity if that is what you were trying to drive at.

wiedergold:
Primary process plus kids

Sent from my Samsung device

-------- Original message --------

jentfoo:
@rusher Yea, I see what you are driving at.  I was nervous to use the already confusing proxy lock more.  But I take it you think we should just use that lock to guard against this condition?

Also, FYI, if a synchronized block ends up not having any thread contention, the JIT optimizes it heavily, and will have very minimal impact (some, in that there will always be a JVM safe point check before it, but very minimal).  This is an example of the types of optimization I mentioned in my email that a ReentrantLock is unable to be optimized for.

jentfoo:
I added a commit to just reuse the `proxy.lock` for now.  Let me know your guys preference, from talking with @rusher I am fine with deferring more locking changes for a bit more.

In case you guys do want to merge this as is (since I forgot to mention it at the start), this entire pull request is submitted under the BSD-new license.

PR #68, opened 2016-01-22
CONJ-248 - fixed infinite recursion in several setters

EGrunzke:
Several methods cause infinite recursion instead of setting parameters.  See also https://mariadb.atlassian.net/browse/CONJ-248


basilmusa:
I just don't get it. How could such a bug exist all that time without being noticed. I mean infinite recursion is not some minor bug. Thanks for your efforts @EGrunzke .

PR #71, opened 2016-02-27
Use the doubleValue of non-BigDecimal Number arguments to setObject.

ingramj:
Previously the longValue was used, which meant that floating point numbers would lose their fractional part. This commit creates a new BigDecimal from the doubleValue instead.


ingramj:
Closing this PR, and opening separate PRs for the two issues.

PR #72, opened 2016-03-03
Improve PacketOutputStream (remove nio, ...)

jacquesmilitello:


src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -1,617 +1,665 @@
...
+import java.io.OutputStream;
+import java.io.Reader;
+import java.io.UnsupportedEncodingException;
 import java.sql.Timestamp;
 import java.util.Calendar;
 import java.util.zip.DeflaterOutputStream;
 
+import org.mariadb.jdbc.internal.util.ByteBuffer;
 
 public class PacketOutputStream extends OutputStream {
```
jentfoo: In my opinion these spacing changes make it VERY hard to evaluate this commit.  If you are going to do spacing changes (though not sure why you would in this case), you should probably do it in a separate commit.

src/main/java/org/mariadb/jdbc/internal/query/MariaDbQuery.java
```diff
@@ -117,8 +118,9 @@ public int writeLastRewritePartLength() {
      */
     public void writeToRewritablePart(final OutputStream os, int rewriteOffset) throws IOException {
```
jentfoo: Why did you not just update the API to accept PacketOutputStream?

src/main/java/org/mariadb/jdbc/internal/util/ByteBuffer.java
```diff
@@ -0,0 +1,271 @@
...
+			}
+		}
+	}
+
+	private void allocate(int size) {
+		this.buffer = new ByteBufferArray(size);
+		this.buffers.add(this.buffer);
+	}
+	
+	private static final class ByteBufferArray {
```
jentfoo: Why did you not just use ByteBuffer?

https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html

It's not obvious what optimization you might be gaining from this that you would not get from just using the existing ByteBuffer.  Using more frequently used classes gives the JIT a better opportunity to optimize it as well (we may or may not get enough usage for the server JIT to actually do some optimizations here).

jentfoo:
Do you actually have any JMH benchmarks or anything else similar to show this is an actual improvement?  I can see how you are trying to avoid memory copies, but I know I personally like to see benchmarks on these kinds of changes to really prove that they are in fact an improvement (things can be quite surprising).

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
jacquesmilitello: Hi Mike,

What do you mean by the "spacing changes" ? the size of the buffer  (1024
-> 4096) ? The target is to have one page size.

Jacques.

src/main/java/org/mariadb/jdbc/internal/query/MariaDbQuery.java
jacquesmilitello: I agree with you, If i can modify the API, I will do it for the next commit.

Thanks

src/main/java/org/mariadb/jdbc/internal/util/ByteBuffer.java
jacquesmilitello: Conceptually we have two kinds of IO (IO and NIO). The mix of the both is
not so good.
Normally, java.nio.byteBuffer is fine for SocketChannel.

The second target is to have two kind for ByteByfferArray (inner class).
The second (ByteBufferInputStream) can be a holder of an inputstream
(remove the copy of the inputStream to memory).

With a nio.bytebuffer we have all the request in memory (file, ....) with a
lot of byteBuffer "copy"

jacquesmilitello:
1/ just he execution of the all test suite (487 junit test) -> before 45.4
(average of 10 executions), after 29.7average of 10 executions).
2/ I can create a JMH test (on the same module or another module ) ? Do you
have a scenario for a test ?
3/ We have tested this change in own solution we have seen (by metrics) an
improvement of 38%.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
vaintroub: that was about whitespace changes in the source files. I agree with mike, it is not possible to see what really changed , becase the diff is in every line

jacquesmilitello: Okay I understand this point.

I will redo the changement ?

vaintroub: I guess it makes sense :)

jacquesmilitello: done !

I have re-apply modification on my last push.

rusher:
Different good idea in this PR.

i'm just wondering about the unsafe part, since JIT must optimize that, and with oracle talking to suppress it. 
JMH test will decide.

jacquesmilitello:
Oracle will not suppress the unsafe part, for JDK9, an alternative will be
available.

src/main/java/org/mariadb/jdbc/internal/util/ByteBuffer.java
jentfoo: I don't see anywhere that you have to copy with a nio ByteBufer that you are able to avoid a copy here?

I understand it is made for NIO originally, but the class is pretty general purpose.

Just my .02, take it for what it is.

jentfoo:
> Oracle will not suppress the unsafe part, for JDK9, an alternative will be available.

Right, but that alternative means that we will now need a version for prior to JDK9, and a version for JDK9 separately.  I don't agree with oracle's decision to further hide it (or even to hide it as much as they do now, as it's a very valuable tool).

But they do make it hard, and it's at least worth considering if it's worth using.  Particularly since it sounds like we will need to maintain two forks once they figure out what the new replacement will be.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -255,66 +225,75 @@ private void checkPacketMaxSize(int limit) throws MaxAllowedPacketException {
...
-
-            if (useCompression) {
-                //now bufferBytes in filled with uncompressed data
-                compressedAndSend(notCompressPosition, bufferBytes);
-            }
-        }
-
-    }
-
+    	buffer.prepare();
```
jentfoo: I will let others comment, but these spacing changes really bother me.  Mostly because it's completely inconsistent with the rest of the project.  Is there a reason you don't want to develop in the already established style?

If it does not bother @rusher though, I am fine with trying to ignore it.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -1,12 +1,17 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Reader;
+import java.io.UnsupportedEncodingException;
 import java.sql.Timestamp;
 import java.util.Calendar;
 import java.util.zip.DeflaterOutputStream;
 
+import org.mariadb.jdbc.internal.util.ByteBuffer;
```
jentfoo: If you are not going to use a NIO ByteBuffer I think you should at minimum rename this class, to avoid confusion.  Anyone who is familiar with java will be confused by this.

src/main/java/org/mariadb/jdbc/internal/util/ByteBuffer.java
jacquesmilitello: 1./ because I use a list of ByteBufferArray and not only one.
2./ I'm working on the next version to wrap directly the inputstream.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -67,7 +59,6 @@ public void startPacket(int seqNo, boolean checkPacketLength) throws IOException
             throw new IOException("Last stream not finished");
         }
         this.seqNo = seqNo;
-        buffer.clear();
```
jentfoo: How are you able to avoid resetting the buffer at these points?  This seems like an odd behavioral change on the surface.

jacquesmilitello:
I agree but a lot of open source API - LIB - products use UNSAFE.
We can have a specific management for JDK9.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
jentfoo: Also, I kind of hate tabs, spaces IME are much more friendly when working with multiple people.

jacquesmilitello: Okay, but where is the eclipse format for the established style ?

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
jacquesmilitello: Ok i will do it for the next commit.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -450,7 +414,6 @@ public PacketOutputStream writeByteArray(final byte[] bytes) {
      * @return this.
      */
     public PacketOutputStream writeByteArrayLength(final byte[] bytes) {
-        assureBufferCapacity(bytes.length + 9);
```
jentfoo: It seems like in some cases this assure capacity could be a performance saver.  If the bytes are larger than your default buffer, and if there is not enough room to write the length, you would allocate for the length write, and then allocate again for the bytes.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -538,7 +495,6 @@ public PacketOutputStream writeFieldLength(long length) {
     public PacketOutputStream writeStringLength(final String str) {
         try {
             final byte[] strBytes = str.getBytes("UTF-8");
-            assureBufferCapacity(strBytes.length + 9);
```
jentfoo: Same as above

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -614,4 +566,8 @@ public PacketOutputStream writeTimeLength(final Calendar calendar, final boolean
         return this;
     }
 
+    public void oneBlock(int size) {
```
jentfoo: Inconsistent spacing

jentfoo:
I am all for the performance boost if it's there and if @rusher and others are okay with the possible maintenance of using Unsafe.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
jentfoo: I am not sure if maria has published one.  But it's very close to the style for my project:

https://github.com/threadly/threadly/blob/master/config/Eclipse_Coding_Style.xml

The biggest change is threadly uses 2 spaces, and mariaDb uses 4.

FWIW, I don't feel every detail of style has to match exactly (others may, not sure).  But the choice to use tabs instead of 4 spaces (or to use a different quantity of spaces) can be the most disruptive visually.

src/main/java/org/mariadb/jdbc/internal/util/ByteBuffer.java
```diff
@@ -0,0 +1,271 @@
...
+		buffer.buf[buffer.pos++] = (byte) x;
+		buffer.buf[buffer.pos++] = (byte) (x >> 8);
+		buffer.buf[buffer.pos++] = (byte) (x >> 16);
+		buffer.buf[buffer.pos++] = (byte) (x >> 24);
+		this.pos += 4;
+	}
+
+
+	public void putLong(long x) {
+		if (this.buffer.pos + 4 > this.buffer.buf.length) {
```
rusher: test size must but 8, not 4

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
rusher: Agree with jentfoo, 
you can launch the following command : 
mvn package -Dmaven.test.skip=true
this will report all code not following rules (like tab indentation)

rusher:
First FTH result not good : a few percent more time.
Additional tests point to the ByteBuffer deletion : 
putInt() was representing 0.6% of the time spend for a query, now it's 3.5%
put(byte) was 0.7%, now is 5,4% ( there is 3 x more call)
put(byte[]) was 0.4%, now is 3.5%

src/test/java/org/mariadb/jdbc/BaseTest.java
```diff
@@ -27,7 +27,7 @@
  */
 @Ignore
 public class BaseTest {
-    protected static final String mDefUrl = "jdbc:mysql://localhost:3306/testj?user=root";
+    protected static final String mDefUrl = "jdbc:mysql://localhost:3306/testj?user=root&password=root";
```
rusher: please remove this, travis test consider having no password. 
for your tests, you can add -DdbURL=jdbc:mysql://localhost:3306/testj?user=root&password=root
that will replace this default string

jacquesmilitello:
How do you mesure this ? FTH ?
How can I reproduce because with my tests the total time is better ?
Thanks in advance.

src/test/java/org/mariadb/jdbc/BaseTest.java
jacquesmilitello: oops, sorry for this commit I will correct it immédiate !!

jentfoo:
> 1/ just he execution of the all test suite (487 junit test) -> before 45.4 (average of 10 executions), after 29.7average of 10 executions).

10 executions is not enough for the JIT to really kick in.  Try 10k executions (which is why JMH is your friend).  You also will want to discard the results during the warm up phase for the most part.

This could be in part why your results don't match @rusher's

jacquesmilitello:
I have JMH and i will create tests to reproduce or do you have jmh tests ?

jentfoo:
> do you have jmh tests?

Sadly I don't for this part of the system.  I am not sure if @rusher has any or not.

src/test/java/org/mariadb/jdbc/BaseTest.java
jacquesmilitello: done in the last commit

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
jacquesmilitello: Done in the last commit (no errors now)

jacquesmilitello:
Hi all,

I have pushed a simple class "org.mariadb.bench.Benchmark" and add a
dependency in the pom to have junit-benchmark.


Can I add JMH test ? (change pom, ... ) ?

Thanks in advance.

src/main/java/org/mariadb/jdbc/internal/stream/PacketOutputStream.java
```diff
@@ -243,124 +257,138 @@ public boolean checkRewritableLength(int length) {
...
+            byte[] header = new byte[HEADER_LENGTH];
+            header[0] = (byte) (length & 0xff);
+            header[1] = (byte) (length >>> 8);
+            header[2] = (byte) (length >>> 16);
+            header[3] = (byte) seqNo++;
+            outputStream.write(header, 0, HEADER_LENGTH);
+            notCompressPosition += 4;
+            
+            if (length > 0) {
+                this.buffer.writeTo(this.outputStream);
```
rusher: must write "length" bytes, like  this.buffer.writeTo(this.outputStream, 0, length);
Packet max size is set in variable maxPacketSize (server option max_allowed_packet). 
if sending more than this size, database will close socket.

rusher:
@acquesmilitello this benchmark can be a good indicator (i may use it to compare perf stat to identify differences), but that must not be in in the test suite, since it doesn't test anything but display some perf result.
I'll check code completely today and launch performance test.

jentfoo:
IMO the benchmark needs to be run TONS more than it currently is.  Set the warmup and rounds to more like `10_000`.  At only 10 + 10, you likely wont be able to take advantage of many JIT optimizations.

rusher:
Perf results show this : 
- current implementation (with ByteBuffer) is more performant for small queries. (around 5% difference). The reason is the methods put* are 3 times faster. I imagine that is because they are intrinsic.
- There is a big improvement for very big query depending on jvm memory, due to the use of direct use of byte array, avoiding allocation of very big ByteBuffer
(I have a lot of memory allocate on JVM, this was why i've not seen the perf difference before)

I don't think we can avoid using ByteBuffer. Since HeapByteBuffer always use array, there may be a solution to use most of the solution anyway. Any other ideas ?

jacquesmilitello:
Hi Diego, with my last commits and push I have optimized small queries. I
have reduced calls of put*.

Can you retry your tests ?

rusher:
Currently PR has missing feature : query more than 16Mb /max_allowed_size don't work anymore. 
So option rewriteBatchStatement isn't working anymore. 
(please run the test case before submitting PR)

With some correction so that works, the FTH tests  show 2 to 10% performance decrease. 
File to reproduced those FTH test (detail results are on the files) :
https://drive.google.com/file/d/0B45mNtoAplKJWFNBQkl6X1VNSWs/view?usp=sharing
https://drive.google.com/file/d/0B45mNtoAplKJM1lwT25udGZ4Vjg/view?usp=sharing

Even if there is some interresting ideas, results are not working accordingly.

jacquesmilitello:
Hi Diego,
 I will fix it as soon as possible.

I have tested with you JMH test and I have these result :

Before :
Benchmark                                         Mode  Cnt       Score
  Error   Units
PerfTextQuery.binaryQuery                        thrpt   50       0,014 ±
  0,001  ops/us
PerfTextQuery.textCharStreamStatement            thrpt   50       0,009 ±
  0,001  ops/us
PerfTextQuery.textQuery                          thrpt   50       0,012 ±
  0,001  ops/us
PerfTextQuery.textRewriteQuery                   thrpt   50      ≈ 10⁻⁵
         ops/us
PerfTextQuery.textRewriteQueryPrepareStatement   thrpt   50      ≈ 10⁻⁴
         ops/us
PerfTextQuery.binaryQuery                         avgt   50      77,288 ±
  8,687   us/op
PerfTextQuery.textCharStreamStatement             avgt   50     100,811 ±
  8,128   us/op
PerfTextQuery.textQuery                           avgt   50      64,260 ±
  6,335   us/op
PerfTextQuery.textRewriteQuery                    avgt   50  203502,943 ±
 2951,505   us/op
PerfTextQuery.textRewriteQueryPrepareStatement    avgt   50    3875,448 ±
594,551   us/op
PerfTextQuery.binaryQuery                       sample   20     109,562 ±
 31,771   us/op
PerfTextQuery.textCharStreamStatement           sample   15     171,972 ±
 53,723   us/op
PerfTextQuery.textQuery                         sample   14      80,571 ±
 62,086   us/op
PerfTextQuery.textRewriteQuery                  sample   50  213233,172 ±
11605,541   us/op
PerfTextQuery.textRewriteQueryPrepareStatement  sample   50    4091,331 ±
636,116   us/op
PerfTextQuery.binaryQuery                           ss   50     129,051 ±
  9,729   us/op
PerfTextQuery.textCharStreamStatement               ss   50     156,878 ±
 17,684   us/op
PerfTextQuery.textQuery                             ss   50     105,272 ±
  7,969   us/op
PerfTextQuery.textRewriteQuery                      ss   50  200260,878 ±
 1629,842   us/op
PerfTextQuery.textRewriteQueryPrepareStatement      ss   50    3825,600 ±
597,972   us/op



after :
Benchmark                                         Mode  Cnt       Score
 Error   Units
PerfTextQuery.binaryQuery                        thrpt   50       0,011 ±
 0,001  ops/us
PerfTextQuery.textCharStreamStatement            thrpt   50       0,012 ±
 0,001  ops/us
PerfTextQuery.textQuery                          thrpt   50       0,015 ±
 0,002  ops/us
PerfTextQuery.textRewriteQuery                   thrpt   50      ≈ 10⁻⁵
        ops/us
PerfTextQuery.textRewriteQueryPrepareStatement   thrpt   50      ≈ 10⁻⁴
        ops/us
PerfTextQuery.binaryQuery                         avgt   50      60,760 ±
 6,349   us/op
PerfTextQuery.textCharStreamStatement             avgt   50      83,004 ±
 5,819   us/op
PerfTextQuery.textQuery                           avgt   50      58,477 ±
 4,681   us/op
PerfTextQuery.textRewriteQuery                    avgt   50  185723,878 ±
3176,932   us/op
PerfTextQuery.textRewriteQueryPrepareStatement    avgt   50    3837,924 ±
 572,227   us/op
PerfTextQuery.binaryQuery                       sample   29      81,024 ±
12,131   us/op
PerfTextQuery.textCharStreamStatement           sample   28     119,803 ±
41,679   us/op
PerfTextQuery.textQuery                         sample   30      69,890 ±
15,652   us/op
PerfTextQuery.textRewriteQuery                  sample   50  186788,086 ±
2692,039   us/op
PerfTextQuery.textRewriteQueryPrepareStatement  sample   50    3928,392 ±
 541,357   us/op
PerfTextQuery.binaryQuery                           ss   50     108,376 ±
 8,575   us/op
PerfTextQuery.textCharStreamStatement               ss   50     147,321 ±
19,660   us/op
PerfTextQuery.textQuery                             ss   50     110,511 ±
 9,670   us/op
PerfTextQuery.textRewriteQuery                      ss   50  185212,025 ±
2979,847   us/op
PerfTextQuery.textRewriteQueryPrepareStatement      ss   50    3835,371 ±
 585,609   us/op


These results are for one Thread.

rusher:
Closing PR, since no performance improvement, and a lot of performance change have been done since

PR #73, opened 2016-03-14
Rethrow connection exceptions after MariaDbServerPreparedStatement fails.

ingramj:
The fix for CONJ-238 was to catch all `SQLException`s thrown by the constructor for `MariaDbServerPreparedStatement`, and fallback to a client prepared statement. This is fine except for `SQLNonTransientConnectionExceptions`, which can be the result of an `IOException` that leaves PacketOutputStream in an invalid state. If that happens, then calling `execute` on the returned `MariaDbClientPreparedStatement` will fail with a mysterious "Last stream not finished" (or, as of 1.3.5, "Stream has already closed") error, with no indication of what previously went wrong. This happens intermittently every few days in our application testing environment.


rusher:
Hi @ingramj 

Good point ! 

Can you please indicate in a comment below that you are contributing your new code of the whole pull request under the BSD-new license or that you have filled out the contribution agreement and sent it.

Thanks,
Diego

ingramj:
Hi Diego,

All code in this PR is contributed under the BSD-new license.

Thanks,

Jim

PR #74, opened 2016-03-14
Use the doubleValue of non-BigDecimal Number arguments to setObject.

ingramj:
Previously the `longValue` was used, which was incompatible with the MySQL Connector/J, and also meant that floating point numbers would lose their fractional part. This commit creates a new `BigDecimal` from the `doubleValue` instead.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/AbstractMariaDbPrepareStatement.java
+++ b/src/main/java/org/mariadb/jdbc/AbstractMariaDbPrepareStatement.java
@@ -844,7 +844,7 @@ public abstract class AbstractMariaDbPrepareStatement extends MariaDbStatement i
                     if (obj instanceof BigDecimal) {
                         setBigDecimal(parameterIndex, (BigDecimal) obj);
                     } else {
-                        setLong(parameterIndex, bd.longValue());
+                        setBigDecimal(parameterIndex, new BigDecimal(bd.doubleValue()));
                     }
                     break;
                 case Types.BIT:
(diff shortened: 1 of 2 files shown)


vaintroub:
I do not think this works for very large long values (say 2^53 or whenever the longs are no more exactly representable as doubles)

ingramj:
Good point. I suppose there should be a type or range check to decide between the double and long values.

vaintroub:
I guess maybe checking the type with instanceof could be the easiest here, as Number does not have too many subclasses. something along the lines below could cut it I guess

if (obj instanceof BigDecimal)
  setBigDecimal(index, obj)
else if (obj instanceof Float  || obj instanceof Double)
  setDouble(index, bd.doubleValue())
else 
  setLong(index, bd.longValue());

rusher:
Hi @ingramj

With this second commit, this is good.

Before validating, like your other PR https://github.com/MariaDB/mariadb-connector-j/pull/73, the MariaDB Foundation needs to have shared ownership of all code that is included in the MariaDB distribution. The easiest way to achieve this is by submitting your code under the BSD-new license. (The other alternative is to sign the code contribution agreement which can be found here: https://mariadb.com/kb/en/mariadb/mca/)

Can you please indicate in a comment below that you are contributing your new code of the whole pull request under the BSD-new license or that you have filled out the contribution agreement and sent it.

Thanks,
Diego

ingramj:
As with the other one, all code in this PR is contributed under the BSD-new license.

Thanks again,

Jim

PR #75, opened 2016-03-18
Performance improvement (mainly buffer)

rusher:
Performance improvement due to :
- using directly the query.getByte("UTF-8") array to avoid recreate a new array.
- rewriteStatement improvement (parsing query to verfy that query can be rewritten + performance) [CONJ-262](https://jira.mariadb.org/projects/CONJ/issues/CONJ-262)
- useCompression correction : local file wasn't working


src/main/java/org/mariadb/jdbc/MariaDbClientPreparedStatement.java
```diff
@@ -160,10 +160,35 @@ public boolean execute() throws SQLException {
...
 
 
+    protected boolean executeInternal() throws SQLException {
+        executing = true;
+        QueryException exception = null;
+        lock.lock();
+        try {
+            executeQueryProlog();
+            batchResultSet = null;
+            Deque<ParameterHolder[]> tmpParam = new ArrayDeque<>(1);
```
jentfoo: FYI, the smallest an ArrayDeque can allocate to is 8, and it must be a 2^x size.  It silently is just upsizing to that minimum size.  Not sure if you want to use a different structure for this now knowing that?  Maybe a Collections.singletonList?

src/main/java/org/mariadb/jdbc/MariaDbClientPreparedStatement.java
```diff
@@ -189,19 +214,91 @@ public void addBatch(final String sql) throws SQLException {
...
+        checkClose();
+        int size = parameterList.size();
+        if (size == 0) {
+            return new int[0];
+        }
+
+        int[] ret = new int[size];
+        int batchQueriesCount = 0;
+        MariaDbResultSet rs = null;
+        cachedResultSets.clear();
```
jentfoo: cachedResultSets is an ArrayDeque, with the use of a lock below it makes me wonder if this structure needs to be synchronized/protected for thread safety?

src/main/java/org/mariadb/jdbc/internal/util/buffer/Buffer.java
```diff
@@ -87,7 +94,18 @@ public String readString(final Charset charset) {
      * @return an short
      */
     public short readShort() {
-        return byteBuffer.getShort();
+        return (short) ((buf[position++] & 0xff)
+                + ((buf[position++] & 0xff) << 8));
```
jentfoo: Does addition make more sense than an or operation?

src/main/java/org/mariadb/jdbc/MariaDbStatement.java
```diff
@@ -1225,20 +1108,21 @@ public void clearBatch() throws SQLException {
         cachedResultSets.clear();
```
jentfoo: Same as above, concerned about thread safety before the lock here

PR #76, opened 2016-03-21
[CONJ-259] Server prepare statement failover use the associated conn…

rusher:
…ection + failover improvement


src/main/java/org/mariadb/jdbc/MariaDbStatement.java
```diff
@@ -154,34 +155,12 @@ public void run() {
         getTimer().schedule(timerTask, queryTimeout * 1000);
```
jentfoo: Why not just set this on one of the other schedulers, or a scheduler that can be provided from the interface we introduced earlier?

If that sort of change interest you at all, I am happy to make a PR.

PR #77, opened 2016-03-22
Pool improvements

jentfoo:
It is probably easier to digest through the commit's, but I figured I would include it all in a single PR.  Feel free to make requests for individual commit adjustments (or removals).

https://github.com/MariaDB/mariadb-connector-j/commit/46c4166b3d8072f5027d9658bc7af4134a842792

This commit is address an issue @rusher pointed me to with tomcat struggling to shutdown.  His recommendation was to track the connections and unscheduled the tasks as needed, and reschedule if we end up creating new connections.

I took a different angle to this.  We have not run into these issues, and I suspect it's because we are using https://github.com/threadly/threadly/ under the hood which defaults to daemon threads.  I am not aware of anything that we would need to hold the app shutdown up for, so it seems like transitioning these threads to daemon is has the least performance impact, and is the easiest option.

https://github.com/MariaDB/mariadb-connector-j/commit/efc6405806147d4910cb75f98ecacb68b882cad8

This one was a slight pet peeve of mine, but @rusher said he would be open to seeing a possible PR solution.  I dislike Timer's because of how they create threads all over the place, and much more prefer pools.  In this implementation we use the existing `DynamicSizedScheduler` to be able to pool these requests.

I considered a trivial implementation where we just increase the pool size in `setTimerTask`, then decreased it in `stopTimeoutTask` if it's set.  But I feel like even though it's a bit more code, it's worth it to support this natively in the `DynamicSizedScheduler`.  This seems the safest, and likely could get reused later on.  In addition by supporting it natively we will be able to get even greater benefit when using our threadly schedulers (it will be super clean, though it would be even cleaner if we just removed all of this and used a threadly "Watchdog" instead, but I understand not wanting to add the dependency).

The final commit https://github.com/MariaDB/mariadb-connector-j/commit/ca0acb238c6bf7df8477806a2f85ae25b58cb0b4 addresses an issue I brought up in PR where this structure is not guaranteed to be locked, and may be accessed concurrently.  I solved those couple conditions and added a few comments where those locked assumptions where being made.

As always, all these changes are submitted under the BSD-new license.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/MariaDbClientPreparedStatement.java
+++ b/src/main/java/org/mariadb/jdbc/MariaDbClientPreparedStatement.java
@@ -234,8 +234,8 @@ public class MariaDbClientPreparedStatement extends AbstractMariaDbPrepareStatem
         int[] ret = new int[size];
         int batchQueriesCount = 0;
         MariaDbResultSet rs = null;
-        cachedResultSets.clear();
         lock.lock();
+        cachedResultSets.clear();
         try {
             if (isRewriteable && (protocol.getOptions().allowMultiQueries || protocol.getOptions().rewriteBatchedStatements)) {
                 batchResultSet = null;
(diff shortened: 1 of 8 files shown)


rusher:
Well done

PR #82, opened 2016-05-18
[CONJ-297] Useless memory consumption when using Statement.setQueryTi…

rusher:
…meout + checkstyle correction

### DIFF
--- a/src/main/java/org/mariadb/jdbc/MariaDbStatement.java
+++ b/src/main/java/org/mariadb/jdbc/MariaDbStatement.java
@@ -49,8 +49,6 @@ OF SUCH DAMAGE.
 
 package org.mariadb.jdbc;
 
-import org.mariadb.jdbc.internal.packet.dao.parameters.ParameterHolder;
-import org.mariadb.jdbc.internal.queryresults.*;
 import org.mariadb.jdbc.internal.protocol.Protocol;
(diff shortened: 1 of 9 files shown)


src/main/java/org/mariadb/jdbc/MariaDbStatement.java
```diff
@@ -49,33 +49,27 @@
...
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import java.util.concurrent.locks.ReentrantLock;
 
 
 public class MariaDbStatement implements Statement, Cloneable {
-    private static final DynamicSizedSchedulerInterface timeoutScheduler =
-            SchedulerServiceProviderHolder.getScheduler(0);
+    //timeout scheduler
+    private static volatile ScheduledThreadPoolExecutor timeoutScheduler = null;
```
jentfoo: Why lazily construct this?  It should not start any threads until it's used anyways.  Seems like it's worth the simpler runtime logic to not lazily construct it.

Also is there any reason not to set this to the interface `ScheduledExecutorService` so that other implementations can be used?

rusher: You're right for the 2 remarks, i will change that now.

PR #85, opened 2016-06-25
CONJ-316: Exception Invalid operation for result set type TYPE_FORWARD_ONLY

leonardlin:
Wrong Exception thrown for ScrollType TYPE_SCROLL_INSENSITIVE: Invalid operation for result set type TYPE_FORWARD_ONLY

When fetchSize is not 0, the mariaDB driver automatically assumes streaming-mode and ignores any scroll-type information set.

This change provides a test to trigger the error and the "hack" solution for it.
It's not ideal but it makes it work.


leonardlin:
I confirm that I contribute the whole pull request under the BSD-new license

rusher:
You're right. 
Thanks for your PR, but as you indicate, this is more a workaround than a correction, so i didn't commit it. 
Issue fixed with https://github.com/MariaDB/mariadb-connector-j/commit/ab052062aef38cc32bf6824f70686632169f2546.
Will be with 1.5 release.

PR #86, opened 2016-07-07
Auto-Discovery Feature for Aurora

catherinep003:
The auto-discovery feature allows users to only have to write the cluster endpoint in their URL and the driver looks for the instance endpoints by itself. Whenever the driver reconnects, such as when a failover occurs, the driver will look for new instances and will remove any unavailable instances. In order to avoid confusion from the users, only the cluster endpoint is accepted in the URL for Aurora. 

Tests were added to verify the functionality of the new feature. 

Some code simplification and typo corrections were done as well.


src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -154,47 +205,238 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
+
+        List<String> currentHosts = new ArrayList<>();
+        for (HostAddress address: addresses) {
+            currentHosts.add(address.host);
+        }
+
+        for (int i = 0; i < addresses.size() && endpoints.size() > 0; i++) {
+            if (!addresses.get(i).host.equals(clusterHost) && !endpoints.contains(addresses.get(i).host)) {
+                removeFromBlacklist(addresses.get(i));
+                addresses.remove(i);
```
jentfoo: This is just an `ArrayList` with no locking.  I am concerned about thread safety here.  I don't see anywhere else this collection is modified, but there seems to be risk in other threads not seeing the updates.

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -154,47 +206,245 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
+     * @throws QueryException
+     */
+    private List<String> getCurrentEndpointIdentifiers(Protocol protocol, String urlEndStr) throws QueryException {
+        List<String> endpoints = new ArrayList<>();
+        try {
+            proxy.lock.lock();
+            try {
+                // Deleted instance may remain in db for 24 hours so ignoring instances that have had no change for IGNORE_TIME_IN_MINUTES
+                Date date = new Date();
+                TimeZone.setDefault(TimeZone.getTimeZone(protocol.getServerData("system_time_zone")));
```
rusher: Default TimeZone musn't be set by driver.

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -114,6 +149,20 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
         resetOldsBlackListHosts();
 
+        // Initial connection to cluster - retrieve endpoints immediately
+        if (searchFilter.isInitialConnection() && urlParser.getHostAddresses().size() == 1 && getClusterHostAddress() != null) {
+            AuroraProtocol tempProtocol = AuroraProtocol.getNewProtocol(getProxy(), getUrlParser());
+            tempProtocol.setHostAddress(getClusterHostAddress());
+            try {
+                tempProtocol.connect();
+                retrieveAllEndpointsAndSet(tempProtocol);
+            } catch (QueryException e) {
```
rusher: On initialConnection, goal is to create connection. 
This newly connection has not to be close: Calling foundActiveMaster(tempProtocol) / foundActiveSecondary(tempProtocol) according to connection type, 
and then checkWaitingConnection() will avoid to search to a new connection afterward.

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -154,47 +206,245 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
+     * @return  the probable host address or null if no valid endpoint found
+     * @throws QueryException
+     */
+    private HostAddress searchForMasterHostAddress(Protocol protocol, List<HostAddress> loopAddress) throws QueryException {
+        String masterHostName = null;
+        proxy.lock.lock();
+        try {
+            Date date = new Date();
+            TimeZone.setDefault(TimeZone.getTimeZone(protocol.getServerData("system_time_zone")));
+            int IGNORE_TIME_IN_MINUTES = 3;
```
rusher: same remark than previous concerning TimeZone.setDefault

src/main/java/org/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -427,13 +427,10 @@ private boolean checkIfPreparable(String sql) {
...
         String cleanSql = sql.toUpperCase().trim();
-        if (cleanSql.contains("SELECT")
+        return cleanSql.contains("SELECT")
                 || cleanSql.contains("UPDATE")
                 || cleanSql.contains("INSERT")
-                || cleanSql.contains("DELETE")) {
-            return true;
-        }
-        return false;
+                || cleanSql.contains("DELETE");
```
zhanhb: REPLACE INTO......

src/main/java/org/mariadb/jdbc/HostAddress.java
```diff
@@ -107,11 +107,16 @@ public HostAddress(String host, int port, String type) {
         String[] tokens = spec.trim().split(",");
         List<HostAddress> arr = new ArrayList<>(tokens.length);
 
-        for (int i = 0; i < tokens.length; i++) {
-            if (tokens[i].startsWith("address=")) {
-                arr.add(parseParameterHostAddress(tokens[i]));
+        // Allow only cluster endpoint to be given unless testing
+        if (haMode == HaMode.AURORA && (tokens.length > 1 || (tokens.length == 1 && spec.indexOf(".cluster-") < 0)) && System.getProperty("auroraFailoverTesting") == null) {
+            throw new IllegalArgumentException("Invalid Aurora connection URL, address should correspond to cluster endpoint");
```
rusher: PR will now throw an exception if there is only one server configured that is not the cluster end-point.
Using cluster-end point now will be documented, but no need to be that hard to users, since this will break there current use.

src/main/java/org/mariadb/jdbc/internal/util/Utils.java
```diff
@@ -444,13 +444,6 @@ public static String nativeSql(String sql, boolean noBackslashEscapes) throws SQ
     public static Protocol retrieveProxy(final UrlParser urlParser, final ReentrantLock lock) throws QueryException, SQLException {
         switch (urlParser.getHaMode()) {
             case AURORA:
```
rusher: Initialy driver permit only master/slave configuration for aurora, but there was different requests to permit having only one server (for integration environnement for example). 
If there is only one instance in cluster, the master-slave implementation will be used, resulting in trying to connect to a slave that will never be there. 
This "mono-instance" case must be handled.

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -154,47 +206,245 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
+                checkWriterAttempts--;
+            } while (currentWriter == null && checkWriterAttempts > 0);
+
+            // Handling special case where no writer is found from secondaryProtocol
+            if (currentWriter == null && getClusterHostAddress() != null) {
+                AuroraProtocol tempProtocol = AuroraProtocol.getNewProtocol(getProxy(), getUrlParser());
+                tempProtocol.setHostAddress(getClusterHostAddress());
+                try {
+                    tempProtocol.connect();
+                    return searchForMasterHostAddress(tempProtocol, loopAddress);
```
rusher: same remark than previous one : goal is to found current master host to directly connect to the current master instance. if master is connected, no need to try a connection after ward, just set current master to this protocol

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
catherinep003: It is only connecting to the cluster endpoint to retrieve the instances for the very first connection. It is being closed because we want to connect to an instance and not the cluster. This logic is only used on initial connection and is only meant to retrieve the instances.

src/main/java/org/mariadb/jdbc/MariaDbConnection.java
catherinep003: Do you mean to add "cleanSql.contains("REPLACE")"? It is not very clear.

src/main/java/org/mariadb/jdbc/HostAddress.java
catherinep003: But keeping the current single instance implementation in Utils would not allow auto-discovery since the self-discovery feature only applies to Aurora specific classes (listener, protocol), more specifically, the code to allow auto-discovery is in AuroraListener and not MastersFailoverListener which is the listener that is instantiated for single endpoint.

src/main/java/org/mariadb/jdbc/MariaDbConnection.java
rusher: that part doesn't concern directly your PR. I've added https://jira.mariadb.org/browse/CONJ-323 to handle this

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
rusher: I understand this logic, but why would we not use this fresh socket, and prefer to wait 1-2 seconds more to create a new connection ? this socket can be use directly. The only difference is that  cluster host can be blacklisted, but since blacklist is only a "connecting order preference" (i.e. driver connect hosts not in blacklist first, and use blacklist after).

src/main/java/org/mariadb/jdbc/HostAddress.java
```diff
@@ -108,8 +110,12 @@ public HostAddress(String host, int port, String type) {
         List<HostAddress> arr = new ArrayList<>(tokens.length);
 
         // Allow only cluster endpoint to be given unless testing
-        if (haMode == HaMode.AURORA && (tokens.length > 1 || (tokens.length == 1 && spec.indexOf(".cluster-") < 0)) && System.getProperty("auroraFailoverTesting") == null) {
-            throw new IllegalArgumentException("Invalid Aurora connection URL, address should correspond to cluster endpoint");
+        if (haMode == HaMode.AURORA && System.getProperty("auroraFailoverTesting") == null) {
+            Pattern clusterPattern = Pattern.compile("(.+)\\.cluster-([a-z0-9]+\\.[a-z0-9\\-]+\\.rds\\.amazonaws\\.com)");
+            Matcher matcher = clusterPattern.matcher(spec);
+            if ((tokens.length > 1 || (tokens.length == 1 && !matcher.find()))) {
+                throw new IllegalArgumentException("Invalid Aurora connection URL, address should correspond to cluster endpoint");
```
rusher: Using only cluster end point will have to be encouraged, but for migration, driver need to permit using existing configuration. All current user of aurora will have this exception thrown.  

A better solution might be having small adjustment to this PR to permit discovery even without using cluster end-point, and since this PR will be merged with the new logging feature. I'll add a warning If  cluster end-point must isn't used

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
catherinep003: I sent you an email with an explanation for it :) Although it references "[...]if master is connected, no need to try a connection after ward, just set current master to this protocol"

src/main/java/org/mariadb/jdbc/HostAddress.java
catherinep003: Is there going to be documentation that will inform the users that using the cluster endpoint is the best scenario? Using the cluster endpoint allows us to handle edge cases that are currently not being handled.

rusher:
Catherine, 

This PR will be added to 1.5 version (beta release for end of july)

Thanks,
Diego

src/main/java/org/mariadb/jdbc/internal/failover/impl/AuroraListener.java
```diff
@@ -154,47 +199,238 @@ public void reconnectFailedConnection(SearchFilter searchFilter) throws QueryExc
...
+    }
+
+    /**
+     * Sets urlParser if there are any changes in the instance endpoints available.
+     *
+     * @param endpoints     instance identifiers
+     * @param clusterHost   host string of the cluster
+     * @param port          port that is common to all endpoints
+     */
+    private void setUrlParserFromEndpoints(List<String> endpoints, String clusterHost, int port) {
```
rusher: Why keeping current cluster end point in the hostAddress list ?

catherinep003: Good point. There is no need since it is being stored in its own variable and used when necessary.

catherinep003:
I am contributing the code under the BSD-new license.

rusher:
closing PR, since merged in Develop branch.

PR #95, opened 2017-01-18
CONJ-412: tinyInt1isBit is not applied in method columnTypeClause

mikecelis:
The method columnTypeClause has been modified to use the tinyInt1isBit flag. This makes sure things such as hibernate validation don't fail. It's solved by checking if the tinyInt1isBit flag is true and when it is true it will add an if to the sql each time the columnName is used to replace any occurences of 'tinyint(1)' with 'BIT'. If the flag is false the columnName will be added unmodified. This has been tested with an application using hibernate-validator-5.2.4.Final and MariaDB 10.1.2

### DIFF
--- a/src/main/java/org/mariadb/jdbc/MariaDbDatabaseMetaData.java
+++ b/src/main/java/org/mariadb/jdbc/MariaDbDatabaseMetaData.java
@@ -86,12 +86,16 @@ public class MariaDbDatabaseMetaData implements DatabaseMetaData {
     }
 
     /* Remove length from column type spec,convert to uppercase,  e.g  bigint(10) unsigned becomes BIGINT UNSIGNED */
-    static String columnTypeClause(String columnName) {
+    String columnTypeClause(String columnName) {
+        String parsedColumnName = parseColumnName(columnName);
+        return " UCASE(IF( " + parsedColumnName + " LIKE '%(%)%', CONCAT(SUBSTRING( " + parsedColumnName + ",1, LOCATE('(',"
+                + parsedColumnName + ") - 1 ), SUBSTRING(" + parsedColumnName + ",1+locate(')'," + parsedColumnName + "))), "
+                + parsedColumnName + "))";
+    }
 
-        return
-                " UCASE(IF( " + columnName + " LIKE '%(%)%', CONCAT(SUBSTRING( " + columnName + ",1, LOCATE('(',"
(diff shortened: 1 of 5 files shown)


src/main/java/org/mariadb/jdbc/MariaDbDatabaseMetaData.java
```diff
@@ -86,12 +86,16 @@ public MariaDbDatabaseMetaData(Connection connection, String user, String url) {
     }
 
     /* Remove length from column type spec,convert to uppercase,  e.g  bigint(10) unsigned becomes BIGINT UNSIGNED */
-    static String columnTypeClause(String columnName) {
+    String columnTypeClause(String columnName) {
+        String parsedColumnName = parseColumnName(columnName);
+        return " UCASE(IF( " + parsedColumnName + " LIKE '%(%)%', CONCAT(SUBSTRING( " + parsedColumnName + ",1, LOCATE('(',"
+                + parsedColumnName + ") - 1 ), SUBSTRING(" + parsedColumnName + ",1+locate(')'," + parsedColumnName + "))), "
+                + parsedColumnName + "))";
+    }
```
rusher: That's correct, but make request very big.
Would be better just to add just another IF, like : 

    static String columnTypeClause(int dataTypeMappingFlags, String columnName) {
        if ((dataTypeMappingFlags  & MariaSelectResultSet.TINYINT1_IS_BIT) > 0) {
            return " UCASE( "
                    + " IF(" + columnName + " = 'tinyint(1)', 'BIT', "
                    + " IF( " + columnName + " LIKE '%(%)%', CONCAT(SUBSTRING( " + columnName + ",1, LOCATE('(',"
                    + columnName + ") - 1 ), SUBSTRING(" + columnName + ",1+locate(')'," + columnName + "))), "
                    + columnName + ")))";
        }
        return " UCASE(IF( " + columnName + " LIKE '%(%)%', CONCAT(SUBSTRING( " + columnName + ",1, LOCATE('(',"
                + columnName + ") - 1 ), SUBSTRING(" + columnName + ",1+locate(')'," + columnName + "))), "
                + columnName + "))";
    }

mikecelis: Okay, I've modified and pushed the code.

rusher:
Hi @mikecelis,

Thanks for you contribution.

Thanks,
diego

mikecelis:
We're contributing the code of this whole pull request under the BSD-new license.

PR #96, opened 2017-01-30
Convert password to bytes using UTF-8 charset

TanyaGaleyev:
While exploring mariadb-connector-j an issue was revealed. If one uses a password which contains non-ascii symbols on Windows platform then he fails to establish jdbc connection using maria-connector-j. In source we found that password is converted to bytes using JVM default charset which on Windows is typically Cp1252. So authentication is failed is a result. If JVM is explicitly started with flag _-Dfile.encoding=UTF8_ then authentication completes successfully.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/util/Utils.java
+++ b/src/main/java/org/mariadb/jdbc/internal/util/Utils.java
@@ -68,6 +68,7 @@ import javax.net.SocketFactory;
 import java.io.IOException;
 import java.lang.reflect.Proxy;
 import java.net.Socket;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.sql.SQLException;
@@ -119,7 +120,7 @@ public class Utils {
         }
 
         final MessageDigest messageDigest = MessageDigest.getInstance("SHA-1");
-        final byte[] stage1 = messageDigest.digest(password.getBytes());
+        final byte[] stage1 = messageDigest.digest(password.getBytes(StandardCharsets.UTF_8));
         messageDigest.reset();
 
         final byte[] stage2 = messageDigest.digest(stage1);


vaintroub:
There is, to my knowledge, unfortunately, no guarantee that server treated password as UTF8, when password was created. I believe it would just take the bytes, it whichever client encoding they came, and store SHA1(SHA1(passwd_bytes)) in the system table . Which means, if passwords were created with the command line client, UTF8-ing them and creating a hash to send to the server is not going to work universally. Which makes password with non-ASCII dependent on environment where they were created. MySQL's Connector/J has passwordCharacterEncoding for that case.

rusher:
Only generic solution is having an option that indicate password encoding (that correspond to MySQL option passwordCharacterEncoding).
That is scheduled in next version ( task is https://jira.mariadb.org/browse/CONJ-212).

TanyaGaleyev:
It is very interesting that in https://jira.mariadb.org/browse/CONJ-212 stated that default value for _passwordCharacterEncoding_ is UTF-8. But currently platform default is used.
Also I have checked same case with MySQL Connector/J and found that is uses UTF-8 encoding by default.
Could you please tell when new version are supposed to be released?

rusher:
touché ! 
I've just changed task description to correct that. 
Next release with this task will be released next week.

TanyaGaleyev:
Great news, thanks!
And another one question to confirm. What will be default value for _passwordCharacterEncoding_ UTF-8 or `Charset.defaultCharset()`?

rusher:
default will be like today : harset.defaultCharset()

rusher:
After some testing, you where right, problem is not the one indicated with https://jira.mariadb.org/browse/CONJ-212 :
 
All exchanges with server are done using utf-8. This charset is set with [client-handshake-response](https://mariadb.com/kb/en/mariadb/1-connecting-connecting/#client-handshake-response) by setting "client character collation" to utf8 / utf8mb4 collation.

Passwords are send using this same [client-handshake-response](https://mariadb.com/kb/en/mariadb/1-connecting-connecting/#client-handshake-response) or by plugin dedicated class if plugin authentication is used. 

Those password always must be encoded using utf-8 for default password authentication send in the same packet "client-handshake-response" - that is this PR - and plugins authentication.

That's actually not the case - getByte() is used - using java default charset may be UTF-8 (unix), CP1250, CP1252 or other depending on OS.

I'm changing https://jira.mariadb.org/browse/CONJ-212, and will correct that.

TanyaGaleyev:
Thanks!

rusher:
Sorry, my tests where somewhat wrong, solution is not this one. 
I'm in conversation with server team to have a better understanding of this issue.

TanyaGaleyev:
Hi @rusher, are there any new on the subject?

TanyaGaleyev:
Actually, for my cases introduction of _passwordCharacterEncoding_ option will be suitable. Then I will make it configurable and set to UTF-8 by default in driver configuration in my application. If password encoding will be UTF-8 by default instead of Charset.defaultCharset() then it will solves the problem too. But I am not 100% sure that it is universal solution. Almost every time we use MySQL server and in tests there were always UTF-8 encoding for passwords on the server side. But was not much experiments with MariaDB server.

rusher:
hi @TanyaGaleyev, 

1.5.9-SNAPSHOT is available with this correction using maven : 
```
<repositories>
    <repository>
        <id>sonatype-nexus-snapshots</id>
        <name>Sonatype Nexus Snapshots</name>
        <url>https://oss.sonatype.org/content/repositories/snapshots</url>
    </repository>
</repositories>

<dependencies>
    <dependency>
        <version>1.5.9-SNAPSHOT</version>
    </dependency>
</dependencies>
```
passwordCharacterEncoding is the only solution : 
There is 2 differents problems concerning password encoding :
- password are crypted on server in some hash based on password current passwords bytes value, that is charset connection dependant (current @@character_set_client).
- java driver password hash is based on String.getBytes(), which is plateform dependant (and local dependant).

Since 5.5, MariaDB use UTF-8 as basis, so this depend on java default charset. Having option passwordCharacterEncoding set to 'UTF-8' will solve this problem. The error message will display current java default charset if this kind of problem occur. 

There has been small issues that will be integrated to 1.5.9 release, that has delayed release of 1.5.9, but it will be tomorrow or the day after.

TanyaGaleyev:
Hey @rusher,

Thanks. Could you please give a hint how to create a user whose password will be encoded to bytes using non UTF-8 character encoding when saving on server.

PR #97, opened 2017-02-04
Update databaseProductName to MariaDB

paladox:


paladox:
This is needed for java programs such as gerrit which use DatabaseMetaData to detect which db class to use. Since we have one that uses the mysql connector but we also want to support the mariadb connector as it will fix our utf8mb4 problems.

rusher:
DatabaseMetaData.getDatabaseProductName() is used a lot to defined specific behaviour for database.
For example hibernate with [StandardDialectResolver](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/engine/jdbc/dialect/internal/StandardDialectResolver.java#L77). 

So ok, it would be better to distinguish "MariaDB" for breaking improvement like ...  i don't know, MariaDB implement a Timestamp with timezone dataType. 
But that's not the case for the moment. But this question will eventually popup

rusher:
closing, since this would break compatibility with existing frameworks

PR #98, opened 2017-02-04
Fix support for jdbc:mariadb in java programs that also support jdbc:…

paladox:
…mysql

It can get confusing in other java programs which support jdbc:mysql for the mysql connector but then want to support jdbc:mariadb for the mariadb connector. This fixes a use case for gerrit where we need to support jdbc:mysql for the mysql connector but want to support jdbc:mariadb to allow us to use session variables to support utf8mb4 and utf8mb4_unicode_gi. Tested locally and fixes it for us. This should not cause any breakages.

This will also be used in conjunction with https://github.com/MariaDB/mariadb-connector-j/pull/97

### DIFF
--- a/src/main/java/org/mariadb/jdbc/UrlParser.java
+++ b/src/main/java/org/mariadb/jdbc/UrlParser.java
@@ -153,7 +153,7 @@ public class UrlParser {
             } else {
                 if (url.startsWith("jdbc:mariadb:")) {
                     UrlParser urlParser = new UrlParser();
-                    parseInternal(urlParser, "jdbc:mysql:" + url.substring(13), prop);
+                    parseInternal(urlParser, url, prop);
                     return urlParser;
                 }
 


rusher:
Hi, this will be taken care in futur 2.0 version : 
see [UrlParser.java](https://github.com/MariaDB/mariadb-connector-j/blob/develop/src/main/java/org/mariadb/jdbc/UrlParser.java#L159)

"jdbc:mariadb" always work, and "jdbc:mysql" work only when connection string doesn't contain "disableMariaDbDriver". This permit having MySQL and MariaDB Driver in the same classpath.

paladox:
Oh thanks

paladox:
@rusher hi, when will 2.0 be released please?

paladox:
@rusher would your patch be able to be back ported for the 1.x release please? It doesn't look all that breaking change to me.

But noticing here https://github.com/MariaDB/mariadb-connector-j/blob/develop/src/main/java/org/mariadb/jdbc/UrlParser.java#L151

mysql was switched to mariadb. So that may cause a problem with mysql. So maybe we want to do it so that if disableMariaDbDriver is passed then it should do mariadb?

rusher:
The goal if this functionnality is to permit having MySQL and MariaDB driver in the same classpath. This isn't possible actually, because when DriverManager ask every registered driver if connection string is handled by driver using [Driver.acceptsURL](https://docs.oracle.com/javase/8/docs/api/java/sql/Driver.html#acceptsURL-java.lang.String-), MariaDB driver actual implementation answer true for "jdbc:mariadb" and "jdbc:mysql". 

So, if MariaDB AND MySQL driver are in the classpath, that will depend on the order of driver registration to DriverManager.

The best would have to use only "jdbc:mariadb" connection string, but that's not possible without breaking a lot of framework (Hibernate, jboss, spark, flyway ...).

So intermediate solution is to explicitly indicated that, connection String that contain "jdbc:mysql" and "disableMariaDbDriver" like  "jdbc:mysql://localhost/db?user=someUser&disableMariaDbDriver" are not handled by driver (https://github.com/MariaDB/mariadb-connector-j/blob/develop/src/main/java/org/mariadb/jdbc/UrlParser.java#L129)

2.0 roadmap is dependant of some server 10.2 task. So, i cannot indicate a date for sure. 
What i can promise is that next version (mid march) will contain this functionnality.

paladox:
@rusher thanks, so your patch will be included in a 1.x release? :)

paladox:
should it be back ported to the master branch for the 1.x release?

davido:
This raises the question: Can I still use this connector without mysql-driver on the class path, but against stock MySQL database? I guess I can, right? Why should I do it, to workaround some session parameters problems that vanilla MySQL-driver doesn't currently seem to respect.

rusher:
@davido. Sure, driver permit to connect MySQL or MariaDB servers, and is self-sufficient. 

Goal is not to delete compatibility at all !

PR #99, opened 2017-02-07
CONJ-426 Allow executeBatch to be interrupted

kevinconaway:
@rusher this PR attempts to address the issue identified in https://jira.mariadb.org/browse/CONJ-426

I believe that the cause of this that [AbstractMultiSend#executeBatchStandard](https://github.com/MariaDB/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/internal/protocol/AbstractMultiSend.java#L271) swallows the `InterruptedException` and does not reset the threads interrupted status.  Thus the caller thread has no idea whether its been interrupted or not.

The `executeBatch` call returns early due to the interrupt but the query is actually running the background.  Thus there is a race condition when querying the state of the `cmdInformation` in `[MariaDbServerPreparedStatement#executeBatch](https://github.com/MariaDB/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/MariaDbServerPreparedStatement.java#L276).  This manifests as the NPE that I linked in the ticket.

My fix in this PR addresses all of this by modifying `AsyncMultiSend` to properly catch the `InterruptedException`, reset the thread interrupt state and rethrow the exception as a `QueryException`.  I also modified this class to cancel the ongoing `AsyncMultiRead` task which is also now interruptible.

I've also modified the data structures in `CmdInformationMultiple` to be thread safe since there is still a tiny chance that the `AsyncMultiRead` task will attempt to write a result while the caller thread in `MariaDbServerPreparedStatement` will attempt to query them.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/queryresults/CmdInformationMultiple.java
+++ b/src/main/java/org/mariadb/jdbc/internal/queryresults/CmdInformationMultiple.java
@@ -53,9 +53,10 @@ import org.mariadb.jdbc.internal.protocol.Protocol;
 import org.mariadb.jdbc.internal.queryresults.resultset.MariaSelectResultSet;
 
 import java.sql.ResultSet;
-import java.sql.SQLException;
 import java.sql.Statement;
-import java.util.*;
+import java.util.Iterator;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 /**
  * Permit to store multiple update count / insert ids.
@@ -70,8 +71,8 @@ import java.util.*;
  */
 public class CmdInformationMultiple implements CmdInformation {
 
-    private Deque<Long> insertIds;
-    private Deque<Integer> updateCounts;
+    private Queue<Long> insertIds;
+    private Queue<Integer> updateCounts;
     private int expectedSize;
 
     /**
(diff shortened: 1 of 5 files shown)


rusher:
Hi @kevinconaway,

Great addition, 

Thanks,
diego

src/main/java/org/mariadb/jdbc/internal/queryresults/CmdInformationMultiple.java
```diff
@@ -168,13 +169,13 @@ public int getCurrentStatNumber() {
 
     @Override
     public boolean moreResults() {
```
jentfoo: FWIW, this could be simplified into just deferring directly to `isCurrentUpdateCount()` since that already handles the `null` check.

kevinconaway:
> Please indicate in a comment below that you are contributing your new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license

Please consider all commits and comments on this work submitted under the BSD-new license.

> The other alternative is to sign the code contribution agreement which can be found here: https://mariadb.com/kb/en/mariadb/mca/

Do I need a username on  on Launchpad / askmonty.org in order to sign this?

kevinconaway:
I've also submitted a copy of the MCA

kevinconaway:
I've modified my test case to only run when `useBatchMultiSend=true`.  The normal protocol uses blocking socket IO which is not interruptible.

kevinconaway:
The build failures linked here https://travis-ci.org/MariaDB/mariadb-connector-j/builds/199340719 seem to be due to ssl / connectivity issues when pulling maven dependencies, not my changes.

rusher:
@kevinconaway, yes, maven/connection on travis did have some random issue yesterday. 
This PR is merge in branch FIX1.5.8 will some change to always have connection in a correct state.

rusher:
closing, since merge in 1.5.8 (https://github.com/MariaDB/mariadb-connector-j/commit/722f7172c1229d02608c872dd4c385e7bd95e548)

PR #102, opened 2017-03-28
Retain leading slash when trust store beings with 'file:///'

pivotal-danjahner:
Removing 8 characters in the case of a file path that starts with 'file:///' removes the leading slash. This causes an error as the connector will assume you have provided a relative file path.


rusher:
Hi @pivotal-danjahner, 

Thanks for you contribution.

Thanks,
diego

pivotal-danjahner:
All of the code in this pull request is being contributed under the BSD-new license.

PR #103, opened 2017-04-07
CONJ-453 fixed to do synchronous io when useBatchMultiSend is false

mwgreen:
Deals with NullPointerException described in CONJ-453.  Appears as if there is no synchronous solution, so we implemented


rusher:
Thanks for having created this issue, this has been corrected in 1.6.0/2.0.0 version.

PR #118, opened 2017-12-05
404 on click  on link

thegounet:
because missing a ' ' before )


rusher:
thanks for the correction.

PR #123, opened 2018-02-15
Transaction is not rollbacked when first statement fails

rPraml:
**Expected behavior**
when a statement fails (e.g. deleting a row which is restricted by a foreign key constraint) a connection.rollback should close the transaction

**Current behavior**
the transaction is still open after `connection.rollback`

**Workaround**
run statement.execute("ROLLBACK") manually

**Possible fix**
There is a test case to demonstrate the issue

In line 117 `sharedConnection.rollback()` effectively does nothing as `protocol.inTransaction()` is false (should be true). See: https://github.com/MariaDB/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/MariaDbConnection.java#L716

`protocol.inTransaction()` is derived from the 'serverStatus' word. Unfortunately, in the previously executed statement we just receive an ERR_PACKET from the server which does not have the serverStatus information.

My suggested fix is, always to send the rollback command to the server


rusher:
Hi @rPraml issue is confirmed. 

I'm still wondering if this is the best way to correct the problem. 
Connection.rollback() and Connection.commit() are intensively used by pool. 
Problem is that the query that throws an error can leave RR lock on the server (that's a bug that I will fill server side - but that's another subject).

When an error is thrown, the server doesn't send an update about server status flag.  A better correction would be that if the last packet from the server was an ERR_Packet, the driver cannot trust server status flag for the transaction then execute commit/rollback.

rusher:
Hi @rPraml ,

Thanks for your contribution!

Thanks,
Diego

rPraml:
Hi @rusher 
all code I will contribute in this PR is under the BSD-new license 

thanks for the quick response and confirming the bug

> I'm still wondering if this is the best way to correct the problem.

It was the minimal invasive way to do this and should not harm (of course, in many cases you do a rollback for nothing, which may affect performance) - the MySql driver behaves similar.
But I agree, that there are better solutions:

I also had the same idea to track a "hadError" flag and do not trust the current server status (this would probably require to fix other places, too)
The best way is probably to add the serverStatus to the ERR-packet struct. But I think this is not so easy, to keep compatibility to other tools / mysql

Roland

rPraml:
Closing this, as it is fixed / merged with [CONJ-586] https://github.com/MariaDB/mariadb-connector-j/commit/aa2bf8b6e7a60a9b84870202c1b97aeca0eb4de7

PR #124, opened 2018-02-16
return a real row number from a result set in case of streaming too

tseylerd:
According to jdbc specification ResultSet.getRow() method must return current row number in all cases except it has TYPE_FORWARD_ONLY type.

It seems like there is no reason to return 0 in case of streaming except the ResultSet has TYPE_FORWARD_ONLY type. If streaming is true and result set has a type different from TYPE_FORWARD_ONLY then the rowPointer always contains real current row number - 1.


rusher:
Hi @tseylerd,

Right, if TYPE_FORWARD_ONLY is not set, the cursor can move backward, so driver keeps the results. row position is then accurate, no reason to return 0 in this case. Thanks for your contribution!
(internal JIRA issue is CONJ-582)

Thanks,
Diego

tseylerd:
I am contributing my new code of the whole pull request, including one or several files that are either new files or modified ones, under the BSD-new license.

PR #129, opened 2018-08-23
AbstractQueryProtocol#isValid optimize code

fgaule:
@rusher i'm assuming galeraAllowableState is inmutable so i can split it values only once instead of doing it on every call.  I have also add 2 log calls when a SocketTimeoutException is raised.

I'm trying to optimize isValid calls after monitoring times in my app.
image

Please feel free to comment and ask for changes.


rusher:
merged on develop branch with commit https://github.com/MariaDB/mariadb-connector-j/commit/1c604d1fa62197281dd1901775d92e4100098815

PR #131, opened 2019-01-11
Made dumpQueriesOnException = false by default as per documentation.

King-hT:
…Added a unit test case for the same.


King-hT:
This PR is raised based on 2 things.
1. Making the code in sync with documentation.
image
2. In our project, we value user data a lot. And there was a security bug reported by the team as the query had the critical user information logged in the stack trace.
Now, this could be a common issue with others where they do not want the queries to be printed on the exception stack trace. 
For now, we have solved this by adding 'dumpQueriesOnException=false' to the host URL.

vaintroub:
If you value security a lot, maybe you should not show the raw SQLException text?

rusher:
Thanks for contribution. this has been [merged](https://github.com/MariaDB/mariadb-connector-j/commit/0294357d5fe24c6122e4bd2ecb0c98d006e84757) on develop branch, with additional commit  https://github.com/MariaDB/mariadb-connector-j/commit/c58c2ab8f2f2801cb7003711c4e601897f347316 for tests correction

PR #132, opened 2019-02-16
[misc] enabled running  of 'SHOW ENGINE INNODB STATUS' for error code 1213

mtykhenko:
- 1213 ER_LOCK_DEADLOCK was not considered to run 'SHOW ENGINE INNODB STATUS'. In this case, lock/deadlock information is also required when corresponding option is enabled from a client-side (includeInnodbStatusInDeadlockExceptions=true)

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/util/exceptions/ExceptionMapper.java
+++ b/src/main/java/org/mariadb/jdbc/internal/util/exceptions/ExceptionMapper.java
@@ -67,7 +67,9 @@ import java.sql.SQLTransactionRollbackException;
 import java.sql.SQLTransientException;
 import java.sql.SQLWarning;
 import java.sql.Statement;
-import java.util.stream.IntStream;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 
 import org.mariadb.jdbc.MariaDbConnection;
 import org.mariadb.jdbc.MariaDbStatement;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mariadb/jdbc/internal/util/exceptions/ExceptionMapper.java
```diff
@@ -132,7 +136,7 @@ public static SQLException getException(SQLException exception, MariaDbConnectio
     if (exception.getSQLState() != null) {
       if (connection != null
           && exception.getSQLState() != null
-          && (1205 == exception.getErrorCode() || 1614 == exception.getErrorCode())) {
+          && (IntStream.of(LOCK_DEADLOCK_ERROR_CODES).anyMatch(lockCode -> lockCode == exception.getErrorCode()))) {
```
rusher: maybe better to use a List definition like 
`private static final List<Integer> LOCK_DEADLOCK_ERROR_CODES = Arrays.asList(1205, 1213, 1614);`
so it can be tested with 
`LOCK_DEADLOCK_ERROR_CODES.contains(exception.getErrorCode())`
avoiding unneeded Stream use

mtykhenko: Agree
Performance is not the main point here but replaced Stream with a Set, because it's more suitable in terms of contains() usage.

PR #134, opened 2019-03-08
[MDEV-15935] support redirection for MariaDB Connector J

GuuBu:
This pull request implements redirection protocol as discussed with community in https://jira.mariadb.org/browse/MDEV-15935.
The basic idea is leveraging OK packet last message to encapsulate the redirection information.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/MariaDbServerCapabilities.java
+++ b/src/main/java/org/mariadb/jdbc/internal/MariaDbServerCapabilities.java
@@ -69,7 +69,7 @@ public class MariaDbServerCapabilities {
   public static final int SSL = 2048; /* Switch to SSL after handshake */
   public static final int IGNORE_SIGPIPE = 4096; /* IGNORE sigpipes */
   public static final int TRANSACTIONS = 8192;
-  public static final int SERVER_SESSION_STATE_CHANGED = 16384; /* Old flag for 4.1 protocol  */
+  public static final int RESERVED = 16384; /* Old flag for 4.1 protocol  */
   public static final int SECURE_CONNECTION = 32768; /* New 4.1 authentication */
   public static final int MULTI_STATEMENTS = 1 << 16; /* Enable/disable multi-stmt support */
   public static final int MULTI_RESULTS = 1 << 17; /* Enable/disable multi-results */
--- a/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
+++ b/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
@@ -73,7 +73,7 @@ public class OkPacket {
    */
   public OkPacket(Buffer buffer, long clientCapabilities) {
 
-    buffer.skipByte(); // fieldCount
+    buffer.skipByte(); // 0x00 OkPacket header
     affectedRows = buffer.getLengthEncodedNumeric();
     insertId = buffer.getLengthEncodedNumeric();
  
--- a/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
+++ b/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
@@ -55,6 +55,7 @@ package org.mariadb.jdbc.internal.com.read;
 import java.nio.charset.StandardCharsets;
 
 import org.mariadb.jdbc.internal.MariaDbServerCapabilities;
+import org.mariadb.jdbc.internal.util.constant.ServerStatus;
 
 public class OkPacket {
 
@@ -76,18 +77,8 @@ public class OkPacket {
     affectedRows = buffer.getLengthEncodedNumeric();
     insertId = buffer.getLengthEncodedNumeric();
  
-    if ((clientCapabilities & MariaDbServerCapabilities.CLIENT_PROTOCOL_41) !=0) {
-    	statusFlags = buffer.readShort();
-    	warnings = buffer.readShort();
-    }
-    else if ((clientCapabilities & MariaDbServerCapabilities.TRANSACTIONS) !=0) {
-    	statusFlags = buffer.readShort();
-    	warnings = 0;
-    } 
-    else {
-    	statusFlags = 0;
-    	warnings = 0;
-    }
+    statusFlags = buffer.readShort();
+    warnings = buffer.readShort();
     
     String message = "";
     String sessionStateMessage = "";
@@ -95,7 +86,7 @@ public class OkPacket {
     	if ((clientCapabilities & MariaDbServerCapabilities.CLIENT_SESSION_TRACK) !=0) {
     		message = buffer.readStringLengthEncoded(StandardCharsets.UTF_8);
     		
-        	if ((statusFlags & MariaDbServerCapabilities.SERVER_SESSION_STATE_CHANGED) !=0)  {
+        	if ((statusFlags & ServerStatus.SERVER_SESSION_STATE_CHANGED) !=0 && buffer.remaining() > 0)  {
         		sessionStateMessage = buffer.readStringLengthEncoded(StandardCharsets.UTF_8);
             }
         } 
--- a/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
+++ b/src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
@@ -3,7 +3,7 @@
  * MariaDB Client for Java
  *
  * Copyright (c) 2012-2014 Monty Program Ab.
- * Copyright (c) 2015-2020 MariaDB Corporation Ab.
+ * Copyright (c) 2015-2019 MariaDB Ab.
  *
  * This library is free software; you can redistribute it and/or modify it under
  * the terms of the GNU Lesser General Public License as published by the Free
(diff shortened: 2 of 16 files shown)


ColinNg:
It seems that the mirror where the AppVeyor Build it gets the MSI packages from, http://mariadb.mirrors.ovh.net/MariaDB/, is missing several versions:

10.2.21 (missing)
10.3.12 (missing)
10.1.37 (missing)
10.0.37 (missing)
5.5.62 (missing)

Whereas these versions are present on the mirror: 
10.3.13 ([present](http://mariadb.mirrors.ovh.net/MariaDB/mariadb-10.3.13/winx64-packages/mariadb-10.3.13-winx64.msi))
10.0.38 (present)
5.5.63 (present)

GuuBu:
> It seems that the mirror where the AppVeyor Build it gets the MSI packages from, http://mariadb.mirrors.ovh.net/MariaDB/, is missing several versions:
> 10.2.21 (missing)
> 10.3.12 (missing)
> 10.1.37 (missing)
> 10.0.37 (missing)
> 5.5.62 (missing)
> Whereas these versions are present on the mirror:
> 10.3.13 (present)
> 10.0.38 (present)
> 5.5.63 (present)

Hi ColinNg，

About your comment, may I ask what action should I take for this to go on further progress? Thanks.

ColinNg:
@GuuBu Probably just need to do a rebase of your branch atop the latest master. Recently master had some commits which has fixed the AppVeyor build failure.

GuuBu:
Hi ColinNg,

Thanks. I've rebased the code and test has been passed. Can you help review the source code change. Thanks.

Best Regards,
Qianqian Bu

vuvova:
Can this be used now? Or is it useless until the server will be able to send redirection information?

GuuBu:
> Can this be used now? Or is it useless until the server will be able to send redirection information?

Hi @vuvova , 
It can be used now with Azure MySQL server. And yes,  it is useless until server will be able to send redirection information. 

Best Regards,
Qianqian Bu

vuvova:
Okay, so Azure MySQL server has it, correct?
Is it your fork of MySQL?
Do you plan to create a pull request for the server part of this functionality?
How the server knows when to send a redirect request and to where?

GuuBu:
@vuvova

Sorry for the late response. Please check the content below. Thanks.

> Okay, so Azure MySQL server has it, correct?
Yes. In order to address customer needs for better latency, we implemented this following a previous discussion with Markus and Vladislav in this Jira thread: https://jira.mariadb.org/browse/MDEV-15935. 

> Is it your fork of MySQL?
We implemented this protocol in the middleware (proxy) and add neccessary support on MySQL server side which is our private repo

> Do you plan to create a pull request for the MariaDB server part of this functionality?
Since the protocol has been discussed through, we can create a pull request to support redirection protocol for MariaDB server.

> How the server knows when to send a redirect request and to where?
As Markus/Vladislav said in the Jira thread, the good part for this protocol is we could leverage existing MySQL protocol of OK Packet for K-V capabilities. Server will always send the redirect info within the OK packet for correct authentications and it would be client driver's choice to reconnect directly to the server.

Best Regards,
Qianqian Bu

src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
```diff
@@ -70,6 +73,11 @@ public OkPacket(Buffer buffer) {
     insertId = buffer.getLengthEncodedNumeric();
     serverStatus = buffer.readShort();
     warnings = buffer.readShort();
+    if (buffer.remaining() > 0) {
+        message = buffer.readStringLengthEncoded(StandardCharsets.UTF_8);
```
bgrainger: According to https://dev.mysql.com/doc/internals/en/packet-OK_Packet.html, if the session supports `CLIENT_SESSION_TRACK` then this is a `string<lenenc>` (length-prefixed string); otherwise, this is `string<EOF>` (string extending from here to the end of the packet).

This code shouldn't assume there's always a length prefix.

src/main/java/org/mariadb/jdbc/internal/util/RedirectionInfo.java
```diff
@@ -0,0 +1,118 @@
...
+    * Parse redirection info from a message return by server.
+    *
+    * @param msg  the string which may contain redirection information.
+    * @return RedirectionInfo host and user for redirection.
+    */
+    public static RedirectionInfo parseRedirectionInfo(String msg) {
+        /**
+         * Get redirected server information contained in OK packet.
+         * Redirection string somehow look like:
+         * Location: mysql://redirectedHostName:redirectedPort/user=redirectedUser
```
vuvova, line 84: A thought. Can this have `/?user=` instead of `/user=` ? That is
```
mysql://redirectedHostName:redirectedPort/?user=redirectedUser
```
that is, to make it a subset of a fairly common jdbc connection string syntax:
```
driver://host:port/database?key=val&key=val&...
```
this will allow to add new features easier without inventing new syntax every time. For example, you specify the redirectedUser here. It's possible that some day someone might want to have redirectedDatabase. Using jdbc url syntax — it's already there, no need to think how to integrate it in the url.

And it'll be easier to explain and document, because it's what people already know.

bgrainger: I, too, thought this syntax was a little strange, especially when `&ttl=NNN` might be appended as another "parameter".

By using `/?user`, any client should be able to use a standard URL-parsing library to extract the parameters, instead of doing manual string parsing. It would also provide a standard answer (i.e., use URL-encoding) for what to do about "special" characters in the username, e.g., how to handle a username containing `&`.

shih-che: @vuvova @bgrainer would you please consider putting comments related to redirection string format to server side PR https://github.com/MariaDB/server/pull/1472 ? As the format is still subject to changes we would like to first change them on server side, and align them to client side after server side PR is completed and the format is settled.

shih-che: I will make the change on server side to use /?user= instead of /user=.

rusher:
Server pull request https://github.com/MariaDB/server/pull/1472 is now OK. 
But we needs a Client implementation to ensure that before merging server PR. 

This PR has been there since more than one year, and connection implementation has been rewritten and simplified since. Could you update PR to current codebase ? I will do that in a few weeks if not.

GuuBu:
Hi @rusher , 
Thanks for reminding, I'll update the PR soon. I notice the OKPacket class is removed. There are several points update according to server side change, and since they are specific to OK packet, can I add the OKPacket class back? 
Thanks.

GuuBu:
Hi @rusher 
I've updated the logic, can you help check?  It seems the test failure is not relevant?

Best Regards,
Qianqian

GuuBu:
I've built a server from PR https://github.com/MariaDB/server/pull/1472, and the basic redirection connection test can pass now.

GuuBu:
Hi @vuvova @bgrainger @rusher 

I've updated the PR according to server side protocol, and add a new option for enableRedirect option. The enableRedirect has three choice now: 
"off" will not use redirection,
 "on" will enforce redirection (will fail if redirection is not available or redirection connection fails), 
"preferred" will use redirection if it is available and the redirection succeed, otherwise, it will just the non-redirected connection.

Could you help check and review? Thanks.

Best Regards,
Qianqian Bu

src/main/java/org/mariadb/jdbc/internal/MariaDbServerCapabilities.java
```diff
@@ -69,7 +69,7 @@
   public static final int SSL = 2048; /* Switch to SSL after handshake */
   public static final int IGNORE_SIGPIPE = 4096; /* IGNORE sigpipes */
   public static final int TRANSACTIONS = 8192;
-  public static final int RESERVED = 16384; /* Old flag for 4.1 protocol  */
+  public static final int SERVER_SESSION_STATE_CHANGED = 16384; /* Old flag for 4.1 protocol  */
```
bgrainger, line 72: This is not the right place to define this flag. It's not part of the "server capabilities" but part of the "server status", and already defined here: https://github.com/mariadb-corporation/mariadb-connector-j/blob/3bc66153b51aca188afc50ff35a0123f16c099ed/src/main/java/org/mariadb/jdbc/internal/util/constant/ServerStatus.java#L69

src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
```diff
@@ -0,0 +1,150 @@
...
+    	warnings = 0;
+    }
+    
+    String message = "";
+    String sessionStateMessage = "";
+    if (buffer.remaining() > 0) {
+    	if ((clientCapabilities & MariaDbServerCapabilities.CLIENT_SESSION_TRACK) !=0) {
+    		message = buffer.readStringLengthEncoded(StandardCharsets.UTF_8);
+    		
+        	if ((statusFlags & MariaDbServerCapabilities.SERVER_SESSION_STATE_CHANGED) !=0)  {
```
bgrainger, line 98:
```suggestion
        	if ((statusFlags & ServerStatus.SERVER_SESSION_STATE_CHANGED) !=0)  {
```

(as per comment about this constant above)

src/main/java/org/mariadb/jdbc/internal/util/RedirectionInfo.java
```diff
@@ -0,0 +1,126 @@
...
+    public String getUser() {
+        return user;
+    }
+    
+    public int getTTL() {
+    	return ttl;
+    }
+
+    /**
+    * Parse redirection info from a message return by server.
```
bgrainger, line 86:
```suggestion
    * Parse redirection info from a message returned by server.
```

src/main/java/org/mariadb/jdbc/internal/util/RedirectionInfo.java
```diff
@@ -0,0 +1,126 @@
...
+         * Location: mysql://[%s]:%u/?user=%s&ttl=%u\n
+         */
+    	String host = "";
+    	String user = "";
+    	int port = -1;
+    	int ttl = -1;
+        try {
+
+            Pattern INFO_PATTERN =
+            		Pattern.compile("^Location: mysql://\\[([^\\]:]+)\\]:([0-9]+)/\\?user=([^&]+)&ttl=([0-9]+)\\n");
```
bgrainger, line 104: Does this need any flexibility for future server-side changes? E.g., allow `ttl` "parameter" to be omitted, or allow additional parameters to be added?

Although, arguably, if additional parameters were added by the server, it might be safest to fail to parse the string on the client, as we can't guarantee that it would safe to simply ignore any extra parameters.

src/main/java/org/mariadb/jdbc/util/DefaultOptions.java
```diff
@@ -493,6 +494,15 @@
           + "deactivating this validation. Hostname verification is disabled when the trustServerCertificate "
           + "option is set",
       false),
+  ENABLE_REDIRECTION(
+      "enableRedirect",
+      RedirectionOption.OFF.toString().toLowerCase(),
+      "2.1.0",
+      "This option let user to decide if use redirection feature and chose different options: "
```
bgrainger, line 501:
```suggestion
      "This option determines how redirection will be used: "
```

src/main/java/org/mariadb/jdbc/util/DefaultOptions.java
```diff
@@ -493,6 +494,15 @@
...
+          + "\n off: Do not use redirection."
+          + "\n on: Must use redirection, fail the connection if server doesnot support redirection or redirection fail."
```
bgrainger, line 503:
```suggestion
          + "\n on: Must use redirection; fail the connection if server does not support redirection or redirection fails."
```

src/main/java/org/mariadb/jdbc/util/DefaultOptions.java
```diff
@@ -493,6 +494,15 @@
...
+          + "\n preferred: Use redirection if redirection and redirection is success, otherwise do not use redirection.",
```
bgrainger, line 504:
```suggestion
          + "\n preferred: Use redirection if supported by the server and redirecting is successful; otherwise, do not use redirection.",
```

src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
```diff
@@ -0,0 +1,150 @@
...
+  private final String sessionStateInfo;
+
+  /**
+   * Read Ok stream result.
+   *
+   * @param buffer current stream's rawBytes
+   */
+  public OkPacket(Buffer buffer, long clientCapabilities) {
+
+    buffer.skipByte(); // fieldCount
```
bgrainger, line 75:
```suggestion
    buffer.skipByte(); // OkPacket Header
```

Does this need to be checked to ensure it's `0x00` (or `0xFE` for EOF payload functioning as OK payload?) or can we just assume it's correct?

src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
```diff
@@ -0,0 +1,150 @@
...
+    affectedRows = buffer.getLengthEncodedNumeric();
+    insertId = buffer.getLengthEncodedNumeric();
+ 
+    if ((clientCapabilities & MariaDbServerCapabilities.CLIENT_PROTOCOL_41) !=0) {
```
bgrainger, line 79: Is this ever false? I.e., does MariaDB Connector/J still support connecting to very old (pre-4.1) servers?

src/main/java/org/mariadb/jdbc/internal/util/RedirectionInfo.java
GuuBu: Has updated the protocol to 
Location: mysql://[%s]:%u/?user=%s&ttl=%u\n

src/main/java/org/mariadb/jdbc/internal/MariaDbServerCapabilities.java
GuuBu: Thanks a lot for your review. I've updated the code.

src/main/java/org/mariadb/jdbc/internal/util/RedirectionInfo.java
GuuBu: From server side's change, the field ttl is always, so I think the code is safe. About the flexibility,  I think maybe we we can handle it in future if new field and related logic is introduced,  right?

src/main/java/org/mariadb/jdbc/internal/com/read/OkPacket.java
GuuBu: I think we can assume it is correct. The ErrorPacket has similar implementation , ang both classes are instantiated only after the first byte is judged.

GuuBu: Has updated the text

rusher:
closing, since implemented in 3.4.0 with https://github.com/mariadb-corporation/mariadb-connector-j/commit/38639ca25800e2f56ebaf8f3dd3a60ca851d1076

PR #135, opened 2019-03-11
Change the query to fetch innodb_read_only from SHOW GLOBAL VARIABLES…

matsuzayaws:
… into SELECT to reduce mutex contention

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/protocol/AbstractConnectProtocol.java
+++ b/src/main/java/org/mariadb/jdbc/internal/protocol/AbstractConnectProtocol.java
@@ -126,7 +126,7 @@ public abstract class AbstractConnectProtocol implements Protocol {
       + "@@system_time_zone,"
       + "@@time_zone,"
       + "@@auto_increment_increment").getBytes(StandardCharsets.UTF_8);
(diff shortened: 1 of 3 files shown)


matsuzayaws:
```fill_variables()``` in ```sql/show.cc```, the function corresponding to ```SHOW GLOBAL VARIABLES``` command, acquires ```LOCK_plugin_delete``` mutex on Aurora MySQL the same as MySQL.

If there are thousands of clients, the client will execute huge amount of ```SHOW GLOBAL VARIABLES LIKE 'innodb_read_only'``` statements and it will then cause performance impact due to mutex contention.  
Since ```SELECT @@innodb_read_only``` doesn't have any side-effect for performance, it is more preferred than ```SHOW GLOBAL VARIABLES```.

rusher:
Need to confirm that is possible. Using "SHOW" was intentional, because Galera detached node only permit SHOW / SET commands.

rusher:
Forget last comment, it doesn't apply in this case. 

Thanks,
diego

matsuzayaws:
Thank you for your comment, Diego. I agree with submitting my code in this PR under the BSD-new license.

PR #136, opened 2019-03-14
[CONJ-695] add AWS RDS IAM authentication

sfcodes:
Added built-in support for **AWS RDS IAM authentication**. It's enabled with the new option `awsIamAuthentication=true` in the URL, and requires AWS SDK RDS 1.x to be on the classpath.

AWS SDK RDS is an _optional_ maven dependency.

For more info about  AWS RDS IAM authentication, see https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html

I documented my thinking about why we need this in
https://jira.mariadb.org/browse/CONJ-695


rusher:
Implementation will be done with a credential pluggable implementation, see  https://github.com/MariaDB/mariadb-connector-j/commit/b653d149dfa5ca378a280609e7cf1982dcd10b43 

Would gladly have any remarks. 

The implementation relies on default AWS SDK DefaultAWSCredentialsProviderChain and DefaultAwsRegionProviderChain, with the possibility to provide options like region by parameters. 
We won't have amazon 2015 root certificates in the driver, that's not driver job to handle amazon certificates

rusher:
closing since AWS IAM authentication (see https://mariadb.com/kb/en/library/about-mariadb-connector-j/#aws-iam)

PR #141, opened 2019-05-20
Should clear AbstractQueryProtocol::localInfileInputStream field after set MariaDbStatement.setLocalInfileInputStream if throw exception

GuuBu:
There are three tests:
StatementTest#testLoadDataInvalidColumn and 
LocalInfileInputStreamTest#testLocalInfileUnValidInterceptor

if the first test is ran before the second test cases, the second test case will fail for sure. 

The reason is that 
StatementTest#testLoadDataInvalidColumn  will set AbstractQueryProtocol::localInfileInputStream field but never clear it.
LocalInfileInputStreamTest#testLocalInfileUnValidInterceptor will use AbstractQueryProtocol::readLocalInfilePacket  in which its logic depends on whether AbstractQueryProtocol::localInfileInputStream is null or not to do following work and decide what to write to table.  Since AbstractQueryProtocol::localInfileInputStream is not cleaned in previous test StatementTest#testLoadDataInvalidColumn, it will use the value and try to write this to target table, if the target table columns are suitable the write will be success, and you will find irrelevant data written to test table.

### DIFF
--- a/src/test/java/org/mariadb/jdbc/StatementTest.java
+++ b/src/test/java/org/mariadb/jdbc/StatementTest.java
@@ -309,6 +309,7 @@ public class StatementTest extends BaseTest {
           fail();
         }
         assertEquals(ER_LOAD_DATA_INVALID_COLUMN_STATE, sqlException.getSQLState());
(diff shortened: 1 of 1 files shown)


rusher:
Reading your description, when stream is send, localInfileInputStream must then be cleared directly by protocol, and that's not the case.

rusher:
After checkin, driver clear internally localInfileInputStream, so ok, that's just if there is any error for better identification

PR #144, opened 2019-06-10
[CONJ-712] RewriteBatchedStatements and useAffectedRows should report no changes correctly

PavelCibulka:
Improvement of https://jira.mariadb.org/browse/CONJ-384

If update of rewritten batch return "no changes" batch return array of "no changes". Otherwise it works as before.

- tested it with our project.  
- test "updateCounts.size() == 1" may be unnecessary


PavelCibulka:
I've created bug report in jira for this:
https://jira.mariadb.org/browse/CONJ-712

PR #151, opened 2020-02-04
[CONJ-759] update conditional to sync currentProtocol to newSecondary on lockAndSwitch

jaredstehler:
This has been running in my environments for a few days now, and it appears to have resolved the issue I was having.

I'm happy to work on adding tests for this, if someone could point me in the right direction?

Fixes https://jira.mariadb.org/browse/CONJ-759


rusher:
This is pushed in develop branch https://github.com/mariadb-corporation/mariadb-connector-j/commit/be2168f939a8d57333b48c056aa6dbcd8fe3d48a and will be part of next corrective release. 
Thanks for pointing that out !

PR #154, opened 2020-03-15
Use java charset caching mechanisms for strings

coold:
String(byte[], int, int, Charset) constructor produces many redundant objects in heap (Decoders/Encoders) due lack of caching.
Using String(byte[], int, int, String) constructor solves this problem.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/MariaDbClob.java
+++ b/src/main/java/org/mariadb/jdbc/MariaDbClob.java
@@ -59,6 +59,7 @@ import java.sql.Clob;
 import java.sql.NClob;
 import java.sql.SQLException;
 import org.mariadb.jdbc.internal.util.exceptions.ExceptionFactory;
+import org.mariadb.jdbc.internal.util.string.StringUtils;
 
 public class MariaDbClob extends MariaDbBlob implements Clob, NClob, Serializable {
 
@@ -95,7 +96,7 @@ public class MariaDbClob extends MariaDbBlob implements Clob, NClob, Serializabl
    * @return string value of blob content.
    */
   public String toString() {
-    return new String(data, offset, length, StandardCharsets.UTF_8);
+    return StringUtils.newString(data, offset, length, StandardCharsets.UTF_8);
   }
 
   /**
@@ -211,7 +212,7 @@ public class MariaDbClob extends MariaDbBlob implements Clob, NClob, Serializabl
    */
   public int setString(long pos, String str) throws SQLException {
     int bytePosition = utf8Position((int) pos - 1);
-    super.setBytes(bytePosition + 1 - offset, str.getBytes(StandardCharsets.UTF_8));
+    super.setBytes(bytePosition + 1 - offset, StringUtils.getBytes(str, StandardCharsets.UTF_8));
     return str.length();
   }
 
--- a/src/main/java/org/mariadb/jdbc/internal/com/read/Buffer.java
+++ b/src/main/java/org/mariadb/jdbc/internal/com/read/Buffer.java
@@ -55,6 +55,7 @@ package org.mariadb.jdbc.internal.com.read;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import org.mariadb.jdbc.internal.util.string.StringUtils;
 
 public class Buffer {
 
@@ -93,7 +94,7 @@ public class Buffer {
     while (remaining() > 0 && (buf[position++] != 0)) {
       cnt++;
     }
-    return new String(buf, initialPosition, cnt, charset);
+    return StringUtils.newString(buf, initialPosition, cnt, charset);
   }
 
   /**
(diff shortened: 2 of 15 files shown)


rusher:
Hi @coold could you explain that assumption ? 
why :
```
try {
   return new String(bytevalue, offset, len, "utf8");
} catch (UnsupportedOperationException u) {
   //handle that
}
```
would be faster compare to use 
```
   return new String(bytevalue, offset, len, StandardCharsets.UTF_8);
```
jmh benchmarks show similar results on jdk8 and better results using Charset for jdk11 ( but that benchmark might induced some java internal caching that i'm not aware). 

Logic would tend to use StandardCharsets.UTF_8 to avoid testing unsupported operation and avoid doing some lookup to load charser encoder.

coold:
Hi @rusher 
Hmm, maybe you are right - they rewritten these classes in java11, but for java8 constructor String(byte[], int, int, Charset) unaware of java.nio.charset.StandardCharsets#UTF_8 and create Decoders and Encoders for every string construction. (but String(byte[], int, int, String) have a thread local cache for known charsets. I know that constructor with checked exception less convenient, but it dont produce unnecessary objects in heap)
I profiled my application with aprof and found that there are too many UTF_8#Decoder in heap (for each string). Don't now about performance though.
Should I write conditional logic at startup to know what java version currently running and then use appropriate method?

coold:
java 8
`    static char[] decode(Charset cs, byte[] ba, int off, int len) {
        // (1)We never cache the "external" cs, the only benefit of creating
        // an additional StringDe/Encoder object to wrap it is to share the
        // de/encode() method. These SD/E objects are short-lifed, the young-gen
        // gc should be able to take care of them well. But the best approash
        // is still not to generate them if not really necessary.
        // (2)The defensive copy of the input byte/char[] has a big performance
        // impact, as well as the outgoing result byte/char[]. Need to do the
        // optimization check of (sm==null && classLoader0==null) for both.
        // (3)getClass().getClassLoader0() is expensive
        // (4)There might be a timing gap in isTrusted setting. getClassLoader0()
        // is only chcked (and then isTrusted gets set) when (SM==null). It is
        // possible that the SM==null for now but then SM is NOT null later
        // when safeTrim() is invoked...the "safe" way to do is to redundant
        // check (... && (isTrusted || SM == null || getClassLoader0())) in trim
        // but it then can be argued that the SM is null when the opertaion
        // is started...
        CharsetDecoder cd = cs.newDecoder();`

java 11:
`    static Result decode(Charset cs, byte[] ba, int off, int len) {
        if (cs == UTF_8) {
            return decodeUTF8(ba, off, len, true);
        }
        if (cs == ISO_8859_1) {
            return decodeLatin1(ba, off, len);
        }
        if (cs == US_ASCII) {
            return decodeASCII(ba, off, len);
        }

        // (1)We never cache the "external" cs, the only benefit of creating
        // an additional StringDe/Encoder object to wrap it is to share the
        // de/encode() method. These SD/E objects are short-lived, the young-gen
        // gc should be able to take care of them well. But the best approach
        // is still not to generate them if not really necessary.
        // (2)The defensive copy of the input byte/char[] has a big performance
        // impact, as well as the outgoing result byte/char[]. Need to do the
        // optimization check of (sm==null && classLoader0==null) for both.
        // (3)There might be a timing gap in isTrusted setting. getClassLoader0()
        // is only checked (and then isTrusted gets set) when (SM==null). It is
        // possible that the SM==null for now but then SM is NOT null later
        // when safeTrim() is invoked...the "safe" way to do is to redundant
        // check (... && (isTrusted || SM == null || getClassLoader0())) in trim
        // but it then can be argued that the SM is null when the operation
        // is started...
        CharsetDecoder cd = cs.newDecoder();`

rusher:
That confirms what one can expect.
I don't know why java 8 need to create a decoder each time (only reason i see i that some charset are statefull ?!), but for StandardCharset this is not really needed. 

So to resume : 
- for java 8 encoding perf are little better using charset name, but difference is not that obvious
- java 11 correct the problem and perf are really better using charset object. 

then let's keep code as currently implemented. Code is more readable / maintainable and is faster in the long run.

coold:
That was not about encoding performance - but about object allocation rate and gc pressure in applications that using this driver. For our small request rate I get 6GB of Decoders in heap after 10 minutes. Less objects - less time we should wait for gc.
I think java create decoders for external Charset to handle case if you implemented your own charset.
But if you reject that - then we should migrate to java11 - to reduce number of objects in heap.

coold:
@rusher do you have any contact? I want to talk with you about another thing that I want to rewrite.

rusher:
you could contact me on diego.dupin@mariadb.com (i'll delete this message just after)

rusher:
>but about object allocation rate and gc pressure in applications that using this driver. For our small request rate I get 6GB of Decoders in heap after 10 minutes. Less objects - less time we should wait for gc.

I may have close this PR too quickly. 

As indicate in java source code : 
>These SD/E objects are short-lifed, the young-gen gc should be able to take care of them well. But the best approash is still not to generate them if not really necessary.

coold:
@rusher What should I do to improve this pull request?

PR #156, opened 2020-07-01
[CONJ-805] >1Byte UTF8 character issue if maxFieldSize is set

MarkusLutum:
Detected at one of my clients.
They use polish special characters like ść
The length in TextRowProtocol is set to 2 for each of those characters so 4 in our example.
With this the current code 
`return new String(buf, pos, Math.min(maxFieldSize * 3, length), StandardCharsets.UTF_8)
                    .substring(0, Math.min(maxFieldSize, length));`
results into an ArrayOutOfBoundsError at the  substring call.

I added a test case and a quick fix for it.
Travis fails on environment issues. Nothing to do with my change.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/com/read/resultset/rowprotocol/TextRowProtocol.java
+++ b/src/main/java/org/mariadb/jdbc/internal/com/read/resultset/rowprotocol/TextRowProtocol.java
@@ -234,8 +234,7 @@ public class TextRowProtocol extends RowProtocol {
     }
 
     if (maxFieldSize > 0) {
-      return new String(buf, pos, Math.min(maxFieldSize * 3, length), StandardCharsets.UTF_8)
-          .substring(0, Math.min(maxFieldSize, length));
+        return new String(buf, pos, Math.min(maxFieldSize, length), StandardCharsets.UTF_8);
     }
 
     return new String(buf, pos, length, StandardCharsets.UTF_8);
--- a/src/test/java/org/mariadb/jdbc/StringUTFTest.java
+++ b/src/test/java/org/mariadb/jdbc/StringUTFTest.java
@@ -1,0 +1,91 @@
+/*
+ *
+ * MariaDB Client for Java
+ *
+ * Copyright (c) 2012-2014 Monty Program Ab.
+ * Copyright (c) 2015-2020 MariaDB Corporation Ab.
+ *
+ * This library is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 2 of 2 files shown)


MarkusLutum:
Files a bug for this as well:
https://jira.mariadb.org/browse/CONJ-805

src/main/java/org/mariadb/jdbc/internal/com/read/resultset/rowprotocol/TextRowProtocol.java
```diff
@@ -234,8 +234,9 @@ public String getInternalString(ColumnDefinition columnInfo, Calendar cal, TimeZ
     }
 
     if (maxFieldSize > 0) {
-      return new String(buf, pos, Math.min(maxFieldSize * 3, length), StandardCharsets.UTF_8)
-          .substring(0, Math.min(maxFieldSize, length));
+    	String value = new String(buf, pos, Math.min(maxFieldSize * 3, length), StandardCharsets.UTF_8);
+    	int maxLength = Math.min(maxFieldSize, length);
+    	return  value.substring(0, Math.min(maxLength, value.length()));
```
rusher, line 239:
```
int maxLength = Math.min(maxFieldSize, length);
return value.substring(0, Math.min(maxLength, value.length()));
```
can be simplified to 

```
if (maxFieldSize < value.length()) {
   return value.substring(0, maxFieldSize);
}
return value;
```
value.length will always be smaller or equal than length.

MarkusLutum: Correct. Changed it to your suggestion.

harawata:
Both the current implementation and the proposed fix count (or try to count) the length of the string, however, the [doc](https://docs.oracle.com/en/java/javase/14/docs/api/java.sql/java/sql/Statement.html#setMaxFieldSize(int)) says...

> Sets the limit for the maximum number of bytes that can be returned for character and binary column values in a ResultSet object produced by this Statement object.

So, it should look something like this, shouldn't it?

```java
if (maxFieldSize > 0) {
  return new String(buf, pos, maxFieldSize, StandardCharsets.UTF_8);
}
```

rusher:
good point @harawata 

In JDBC from [spec ](https://download.oracle.com/otndocs/jcp/jdbc-4_3-mrel3-spec/index.html):

> 8.3.1 Silent Truncation
> The Statement.setMaxFieldSize method allows a maximum size (in bytes) to
> be set. This limit applies only to the BINARY, VARBINARY, LONGVARBINARY, CHAR,
> VARCHAR, LONGVARCHAR, NCHAR, NVARCHAR, and LONGNVARCHAR data types.
> If a limit has been set using setMaxFieldSize and there is an attempt to read data
> that exceeds the limit, any truncation that occurs as a result of exceeding the set limit
> will not be reported.

CHAR and VARCHAR are explicitly listed.

This is then not just a bug correction, but a change of behavior compares to current driver implementation, so has then to be in the next 2.7 version, not in a corrective patch version.

MarkusLutum:
So what to do @rusher. Should I quickly change my pull request to @harawata s suggestion? And you do the rest?

PR #157, opened 2020-08-05
Fix - setConfiguration not being called on classes that extend ConfigurableSocketFactory

shahmirn:
In my project, I have a class that extends ConfigurableSocketFactory, but the setConfiguration method is never called. I added a breakpoint to if (socketFactoryClass.isInstance(ConfigurableSocketFactory.class)) {, and the expression was evaluating to false.

From my googling, I landed at https://stackoverflow.com/questions/31083699/java-class-isinstance-always-returns-false, which indicated that the evaluation was being done backwards.

When I evaluated ConfigurableSocketFactory.class.isInstance(socketFactory), the result was true


rusher:
Thanks for detail and fix 👍 

Jira task created https://jira.mariadb.org/browse/CONJ-821 
fix is merged into develop branch ( https://github.com/mariadb-corporation/mariadb-connector-j/commit/a88d392fb5eb6ae32cf83ceea16bc33df03e1265) so will be part of 2.6.3 release

PR #158, opened 2020-08-11
 since array[3] will be uses as a wholepart, in case there will be / in this part, e.g. a linux ca path, it should limit split num

GuuBu:
since array[3] will be uses as a wholepart, in case there will be / in this part, e.g. a linux ca path, it should limit split num

### DIFF
--- a/src/test/java/org/mariadb/jdbc/BaseTest.java
+++ b/src/test/java/org/mariadb/jdbc/BaseTest.java
@@ -528,7 +528,7 @@ public class BaseTest {
     } catch (IOException e) {
       e.printStackTrace();
     }
-    String[] splitValue = connUri.split("/");
+    String[] splitValue = connUri.split("/", 4);
     String[] subarray =
         Arrays.asList(splitValue).subList(3, splitValue.length).toArray(new String[0]);
     String dbAndParameters = String.join("/", subarray);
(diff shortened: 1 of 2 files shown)


src/test/java/org/mariadb/jdbc/BaseTest.java
```diff
@@ -528,7 +528,7 @@ public Connection createProxyConnection(Properties info) throws SQLException {
     } catch (IOException e) {
       e.printStackTrace();
     }
-    String[] splitValue = connUri.split("/");
+    String[] splitValue = connUri.split("/", 4);
```
rusher, line 531: I don't see the point of this change, this is test code only and because if there is more slash, remaining values are joined by the next lines :  
```
 String[] subarray =
        Arrays.asList(splitValue).subList(3, splitValue.length).toArray(new String[0]);	        Arrays.asList(splitValue).subList(3, splitValue.length).toArray(new String[0]);
    String dbAndParameters = String.join("/", subarray);	    String dbAndParameters = String.join("/", subarray);
```
but sure, the whole part 
```
    String[] splitValue = connUri.split("/");
    String[] subarray =
        Arrays.asList(splitValue).subList(3, splitValue.length).toArray(new String[0]);
    String dbAndParameters = String.join("/", subarray);
```
can be simplified by 
```
    String[] splitValue = connUri.split("/", 4);
    String dbAndParameters = splitValue.length == 4 ? splitValue[3] : "";
```

GuuBu: Hi @rusher ,
Yes, sorry I didn't read the updated left part carefully. It should be no problem then. I'll close the PR.

PR #164, opened 2020-10-28
8.0 will use the foreign key name as constraint name

GuuBu:
From test, mysql 8.0 will use the foreign key name as constraint name, so the test condition should also exclude 8.0
Tested version: 8.0.15


GuuBu:
reverted in 8.0.16

PR #168, opened 2021-03-07
BinaryRowProtocol: redundant objects creation while reading data from BIGINT column

AmigoV:
Now while getting `long` value from unsigned `BIGINT` column, method `getInternalLong`, unnecessary objects will be created: two `BigInteger` and one `String`. Also, redundant conversion from `byte array` to `long` and back from `long` to `byte array` will be performed. The same can be partly applied to `getInternalFloat`, `getInternalDouble`, and `getInternalBigDecimal`.

There is a more straightforward way to check that a value is in Long range that doesn't require additional objects creation. 
Also, if a column is unsigned, there is no need to convert an array of bytes to `long` just to convert that value back to bytes to create `BigInteger`.
`BigDecimal` can be created directly from `BigInteger` - no need to convert `BigInteger` to a string.


rusher:
Clearly. this will be part of next release.

rusher:
closing since merged in 2.7.3

PR #170, opened 2021-03-16
Fixed java.lang.NoClassDefFoundError: javax/sql/rowset/serial/SerialException when using OSGi framework

m-rm:
We are using Apache Karaf and recently upgraded pax-jdbc-mariadb from version 1.4.5 to 1.5.0. This increased the version of mariadb-java-client from 2.4.4 to 2.7.2. When connecting to a database the bundle crashed with `java.lang.NoClassDefFoundError: javax/sql/rowset/serial/SerialException`. The solution is to add a missing import to pom.xml.


rusher:
closing since merged in 2.7.3

PR #171, opened 2021-05-17
Source code repo tag inconsistent with maven repo 3.0.0 vs 3.0.0-alpha

GuuBu:
Do not find a way to raise issue, so create a PR instead.
The tag in code repo is 3.0.0 while the folder name under https://repo.maven.apache.org/maven2/org/mariadb/jdbc/mariadb-java-client/    is 3.0.0-alpha. Which should be an inconsistence issue. This can break some automatic package pull operation. Can we make the tags consistent?


rusher:
Hi @GuuBu 
I've corrected tag to 3.0.0-alpha. 
thanks for reporting that

PR #172, opened 2021-05-26
CONJ-885 Log any SQLException that occurs while adding connections to pool

jaikiran:
The commit here fixes the issue noted in https://jira.mariadb.org/browse/CONJ-885 where if a `SQLException` occurs while adding a new connection to the pool, then that exception is never reported/logged which can then make it difficult to understand subsequent failures to obtain connection(s) from the `Pool` instance.


jaikiran:
The failure in travis CI doesn't look related to this change and instead appear to be environment issues:

```
ERROR: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

jaikiran:
Hello @rusher, any help reviewing this please?

PR #173, opened 2021-08-24
CONJ-896 Fix regression in pooled connections

jaikiran:
Hello @rusher, I think the commit here should address the issue noted in https://jira.mariadb.org/browse/CONJ-896.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/internal/util/pool/Pool.java
+++ b/src/main/java/org/mariadb/jdbc/internal/util/pool/Pool.java
@@ -375,7 +375,7 @@ public class Pool implements AutoCloseable, PoolMBean {
 
             // ensure that other connection will be validated before being use
             // since one connection failed, better to assume the other might as well
-            idleConnections.stream().forEach(c -> c.lastUsedToNow());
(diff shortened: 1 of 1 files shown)


jaikiran:
The travis job failures don't look related to this change. Logs say:

```
CONNECTOR_TEST_SECRET_KEY env not set

{"ipv6":true,"fixed-cidr-v6":"2001:db8:1::/64"}

private environment variable CONNECTOR_TEST_SECRET_KEY must be provided for skysql-ha
```
```
CONNECTOR_TEST_SECRET_KEY env not set

{"ipv6":true,"fixed-cidr-v6":"2001:db8:1::/64"}

private environment variable CONNECTOR_TEST_SECRET_KEY must be provided for skysql
```

rusher:
Thanks for your contribution.
The problems concerning test failing have been corrected with https://github.com/mariadb-corporation/mariadb-connector-j/commit/06852c1eb3c7e360520f303b229bfc47deac91d5 (that was affecting PR only)

jaikiran:
Thank you for reviewing this and merging @rusher

PR #182, opened 2023-01-11
Fix broken DatabaseMetaData.getColumns() on older server versions

eirikbakke:
On older versions of MariaDB and MySQL, DatabaseMetaData.getColumns() will fail with the error `Unknown column 'DATETIME_PRECISION' in 'field list'`. This is due to the `DATETIME_PRECISION` column not being present in the `INFORMATION_SCHEMA.COLUMNS` table. This column was introduced as part of the introduction of microsecond precision in TIME/DATETIME/TIMESTAMP types, in [MariaDB 5.3.0](https://web.archive.org/web/20130928042640/https://mariadb.com/kb/en/microseconds-in-mariadb) (I believe) and [MySQL 5.6.4](https://dev.mysql.com/doc/relnotes/mysql/5.6/en/news-5-6-4.html).

The proposed fix adjusts the metadata query for the relevant older server versions. The fixed behavior in these cases is as if the DATETIME_PRECISION field had a value of 0, which is consistent with the fact that these older database versions did not support any fractional seconds digits.

Similar [conditional logic](https://github.com/mysql/mysql-connector-j/blob/dd61577595edad45c398af508cf91ad26fc4144f/src/main/user-impl/java/com/mysql/cj/jdbc/DatabaseMetaDataUsingInfoSchema.java#L158) already exist in the MySQL project's JDBC driver.

(This bug was [reported](https://github.com/ultorg/public_issues/issues/47) by a user of [Ultorg](https://www.ultorg.com), a graphical database client. To test the modified MariaDB JDBC driver, I ran Ultorg's integration test suite with it while connecting to MariaDB 10.6.11, MySQL 5.5.62, and MySQL 8.0.26.)

### DIFF
--- a/src/main/java/org/mariadb/jdbc/DatabaseMetaData.java
+++ b/src/main/java/org/mariadb/jdbc/DatabaseMetaData.java
@@ -9,6 +9,7 @@ import java.sql.Statement;
 import java.text.ParseException;
 import java.util.*;
 import org.mariadb.jdbc.client.DataType;
+import org.mariadb.jdbc.client.ServerVersion;
 import org.mariadb.jdbc.client.result.CompleteResult;
 import org.mariadb.jdbc.client.result.Result;
 import org.mariadb.jdbc.util.VersionFactory;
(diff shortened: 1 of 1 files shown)


rusher:
Thanks for the PR. 
Those version are not supported anymore, but no reason to add correction for them. 
merged on develop branch. Will be on 3.1.2 release

eirikbakke:
Thank you for the quick review and merge! (Handling some older unsupported server versions is useful for tools that are supposed to be able to connect to arbitrary MySQL/MariaDB instances, and lets us use the MariaDB driver for both.)

PR #186, opened 2023-06-27
fix readonly connection set 

lure:
According to the documentation, MariaDB support the `read-only` mode. 
However, having the `Connection.isReadOnly == true` results in  `stateFlag = 28` which contains binary `100`  of ConnectionState.STATE_READ_ONLY, while database reports the TX as `trx_is_read_only: 0`


rusher:
Sorry, i don't understand either the description. 

proposal only will ensure calling setReadOnly even when not required, and with wrong value : after reset, connection must behave like a newly created connection, so must have read only set to false. Current implementation is then accurate.

PR #190, opened 2024-01-02
Fix NullPointer in BasePreparedStatement#setURL

nicolube:


nicolube:
... Some kind of limit was reached:

```
: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

rusher:
Thanks a lot for correction !

PR #196, opened 2024-08-18
Socket connections now try all possible IP addresses

dreamlike-ocean:
Fixed an issue in ConnectionHelper where Socket connections sometimes failed. Socket connections now try all possible IP addresses, not just the first one. 

When a host has multiple IP resolution results, especially when the results include both IPv4 and IPv6 addresses, this change can help connect to the correct IP. 

For example, if the server is listening on ::1 and the client uses `localhost` to connect, by default, it will only attempt to connect to 127.0.0.1


rusher:
At some point, i would prefer "Happy Eyeballs v2" ([RFC 8305](https://www.rfc-editor.org/rfc/rfc8305.txt)) when implemented in java : application that uses a Happy Eyeballs algorithm checks both IPv4 and IPv6 connectivity (with a preference for IPv6) and uses the first connection that connects successfully.

PR #198, opened 2024-11-19
Fixing flaky configuration tests

mihirgune:
## Description

This Pull Request is intend to fix flakiness in the following tests in [org.mariadb.jdbc.unit.util.ConfigurationTest](https://github.com/mariadb-corporation/mariadb-connector-j/blob/9c2db3bee043c71755a35ace70fde10c8fffa6a9/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java#L20)

1. [org.mariadb.jdbc.unit.util.ConfigurationTest#testBuild](https://github.com/mariadb-corporation/mariadb-connector-j/blob/9c2db3bee043c71755a35ace70fde10c8fffa6a9/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java#L81)
2. [org.mariadb.jdbc.unit.util.ConfigurationTest#builder](https://github.com/mariadb-corporation/mariadb-connector-j/blob/9c2db3bee043c71755a35ace70fde10c8fffa6a9/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java#L869)
3. [org.mariadb.jdbc.unit.util.ConfigurationTest#toConf](https://github.com/mariadb-corporation/mariadb-connector-j/blob/9c2db3bee043c71755a35ace70fde10c8fffa6a9/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java#L1016)

These tests fail intermittently when executed with the [NonDex](https://github.com/TestingResearchIllinois/NonDex) plugin due to non-deterministic ordering of properties. The failure logs are attached for reference:

- [ConfigurationTest_java_nondex_output.log](https://github.com/user-attachments/files/17807624/ConfigurationTest_java_nondex_output.log)

### Cause of the Issue:

The failures occur because the tests compare expected strings with actual strings generated from the `Configuration` object. The ordering of query parameters and non-mapped options in these strings is non-deterministic due to the use of `java.util.Properties`, which does not guarantee order. This leads to mismatches in the expected and actual outputs during assertions.

### Fix

All three tests (`testBuild`, `builder`, and `toConf`) fail because they perform assertions comparing expected strings to actual strings generated from the `Configuration` object. However, the ordering of query parameters in the connection string is non-deterministic due to the use of `java.util.Properties`, which does not guarantee the order of its elements. This nondeterminism leads to mismatches during assertions, causing the tests to fail.

To resolve this issue for the first two tests (`testBuild` and `builder`), we introduced a new method called `normalizeConnectionString()`. This method parses the connection string, extracts the query parameters, sorts them lexicographically, and reconstructs the string with the sorted parameters. By fixing the order of the query parameters, these tests now pass consistently. For the third test (`toConf`), we addressed the same issue by introducing a new method called `normalizeConfigurationString()`. This method ensures that `normalizeConnectionString()` is called on the URL generated within the configuration that the `toConf` test checks, thereby standardizing the ordering of query parameters and allowing the test to pass reliably.

In addition to the above issue, the `toConf` test fails occasionally because the `conf.nonMappedOptions` (unknown options) in `org.mariadb.jdbc.Configuration#toConf` are unordered. Although the method ensures that other option types (non-default options, default options) are ordered, the unordered `nonMappedOptions` lead to inconsistent outputs during assertions, causing the test to fail. To fix this, we modified the `toConf` method to sort the `conf.nonMappedOptions`, ensuring consistent ordering and eliminating the flakiness in the `toConf` test.

PS: Instead of sorting the `conf.nonMappedOptions` in `org.mariadb.jdbc.Configuration#toConf`, I can also implement a fix wherein the test is made flexible to accept a configuration string that contains the desired options in any order.

Please let me know if any other changes are required. Thanks!

### DIFF
--- a/src/main/java/org/mariadb/jdbc/Configuration.java
+++ b/src/main/java/org/mariadb/jdbc/Configuration.java
@@ -1039,9 +1039,13 @@ public class Configuration {
     if (conf.nonMappedOptions.isEmpty()) {
       sbUnknownOpts.append("None");
     } else {
-      for (Map.Entry<Object, Object> entry : conf.nonMappedOptions.entrySet()) {
-        sbUnknownOpts.append("\n * ").append(entry.getKey()).append(" : ").append(entry.getValue());
-      }
+      conf.nonMappedOptions.entrySet().stream()
+              .map(entry -> new AbstractMap.SimpleEntry<>(
+                      entry.getKey().toString(),
+                      entry.getValue() != null ? entry.getValue().toString() : ""
+              ))
+              .sorted(Map.Entry.comparingByKey())
+              .forEach(entry -> sbUnknownOpts.append("\n * ").append(entry.getKey()).append(" : ").append(entry.getValue()));
     }
     sb.append("Configuration:")
         .append("\n * resulting Url : ")
--- a/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java
+++ b/src/test/java/org/mariadb/jdbc/unit/util/ConfigurationTest.java
@@ -7,7 +7,11 @@ import static org.junit.jupiter.api.Assertions.*;
 
 import java.sql.DriverManager;
 import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Properties;
+import java.util.stream.Collectors;
+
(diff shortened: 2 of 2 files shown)


rusher:
thanks for PR, merged

PR #203, opened 2025-05-07
Speed up concurrent pool creation

emaccaull:
Creating many pools concurrently was slow because all threads are contending for `poolMap`, which is held while a `Pool` is being initialized (which can take a few seconds).

PoolHolder was introduced to lazily create a Pools. Creating a pool instance is synchronized on the PoolHolder and therefore only requests for the same pool instance (referenced by Configuration) will face contention.

Creating many Pool instances concurrently is probably not a usual use case, but does affect sharded environments where shards are behind different databases hosts. In my particular case, I need to run cross-shard queries to understand production data, which is a debugging and business insights task.


emaccaull:
Closing for now. Will create a new PR with Travis working.

PR #204, opened 2025-05-12
Speed up concurrent pool creation

emaccaull:
Initializing multiple pools in parallel incurred significant lock contention on `poolMap` because this lock was held while any `Pool` was being constructed (which could take several seconds).

Because of this, connecting to independent databases in parallel was not possible (Pool creation was serialized), severely limiting overall performance.

## Changes

* Add `PoolHolder` to defer creation of `Pool` and act as a lock for each pool key.
* Insert a new `PoolHolder` while holding the `poolMap` lock, reducing global thread contention.
* Threads looking up the same key will acquire the same `PoolHolder` and that `PoolHolder` will ensure that only one `Pool` is created. Threads looking up unrelated `PoolHolder` will not be blocked by `Pool` creation.
* Make registerJxm/unRegisterJmx thread safe. When Pool creation happened in parallel, there were errors in the logs because `Pool` was attempted to be registered more than one.

## Testing

This change should have no impact on functionality. Performance should be improved when creating many Pools in parallel. Performance should remain the same when creating `Pool` instances sequentially.

* Unit test added to validate that many pools can be created in parallel without issue.
* Tested against ~20 real DB servers and saw performance improvement. All connections were created at the same time on separate threads.
* Ran all unit tests with Travis (see https://github.com/emaccaull/mariadb-connector-j/pull/1).

### DIFF
--- a/src/main/java/org/mariadb/jdbc/pool/Pool.java
+++ b/src/main/java/org/mariadb/jdbc/pool/Pool.java
@@ -534,8 +534,10 @@ public class Pool implements AutoCloseable, PoolMBean {
     String jmxName = poolTag.replace(":", "_");
     ObjectName name = new ObjectName("org.mariadb.jdbc.pool:type=" + jmxName);
 
-    if (!mbs.isRegistered(name)) {
-      mbs.registerMBean(this, name);
+    synchronized (mbs) {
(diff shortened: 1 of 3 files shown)


rusher:
nice addition. thanks

PR #208, opened 2025-09-03
[feat] add defaultAuth for MariaDB JDBC which is simular to cli default-auth

JackyWuSJM:
Add defaultAuth for MariaDB JDBC which is similar to cli default-auth.


rusher:
I don't really know what the point of this  PR : mysql/mariadb protocol send in initial handshake packet  https://mariadb.com/docs/server/reference/clientserver-protocol/1-connecting/connection#initial-handshake-packet authentication plugin name that indicates the default authentication plugin to use. 

it makes sense to rely on that value. MariaDB stil always send native password value, but https://jira.mariadb.org/browse/MDEV-12320 will be done at some point.

rusher:
closing since no answer

PR #213, opened 2026-03-03
Fix: replace [--] with [-] in LOAD LOCAL INFILE regex patterns

sebdomdev:
[--] is a degenerate character range (start == end) that is not portable across regex engines. [-] unambiguously matches a literal hyphen and is the correct way to express a single-character class containing only -.


rusher:
released in 3.5.9

PR #214, opened 2026-03-10
fuzz: add professional fuzzing suite and maven infrastructure

vishalcoc44:
This introduces a native, coverage-guided fuzzing suite (powered by Jazzer) directly into the repository. It's designed to automatically hunt down edge cases, memory leaks, and protocol parsing bugs without you having to write a million manual unit tests. 
To keep things perfectly clean, all fuzzing logic is isolated in `src/fuzz/java` and runs entirely under a dedicated `fuzz` Maven profile.
We need this because manually predicting every malformed byte a server might throw at the driver is basically not possible only.
This setup has already pushed fuzzing coverage from a baseline of 12 lines( which I checked on the fuzz introspector report) to over **858 internal counters** (a 71.5x increase). 
In local testing, this exact suite successfully flushed out some bugs hidden in the packet decoding logic (I'll be submitting those bug reports separately

* **The Fuzzers:** Deep-dive fuzzers targeting URL parsing, the binary protocol, and `ColumnDecoder` stress-testing.
* **The Mocks:** A unified, high-fidelity mock context (`FuzzContext.java`) so we don't have to spin up an actual database socket just to test driver logic.
* **The Build:** A new `<profile>` in `pom.xml`.
* **Zero Disruption:** Standard builds (`mvn clean install`) are completely unaffected. You have to explicitly ask Maven to run the fuzzers (`mvn compile -Pfuzz`).

if we can get this merged here, we can edit the google oss fuzz repo to point towards this area of fuzzers to do the fuzzing.
I'd really like any kind of input.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,12 @@ jobs:
             if mysql -h localhost -u root -p"${{ env.TEST_DB_PASSWORD }}" -e "SELECT 1;" > /dev/null 2>&1; then
               echo "MariaDB is up"
               # Verify if the problematic variable exists and handle it if it does
-              mysql -h localhost -u root -p"${{ env.TEST_DB_PASSWORD }}" -e "SET @@simple_password_check_digits = IF(EXISTS(SELECT 1 FROM information_schema.SYSTEM_VARIABLES WHERE VARIABLE_NAME='simple_password_check_digits'), @@simple_password_check_digits, 0);" || echo "Skipping variable set"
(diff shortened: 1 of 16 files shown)


vishalcoc44:
All stable versions pass 
All CI checks for stable MariaDB versions pass successfully:

 MariaDB 10.6, 10.11, 11.4 (Java 8, 11, 17, 21, 25), 11.8, 12.0, 12.1, 12.2
The only failure is MariaDB 12.3 dev, which is a pre-release/development version. The failure is in FailoverTest.transactionReplayPreparedStatement() which seems to be a known flaky test that fails intermittently due to CI network timing during failover proxy simulations. This is not related to this PR.
The last successful main build that happened for this repo also confirms the fact thatci setup treats it as non fatal.
@rusher could you check this out

vishalcoc44:
@rusher

PR #217, opened 2026-05-26
use Locale.ROOT in DurationCodec text formatting

jmestwa-coder:
A Duration bound on a text-protocol statement is formatted with String.format under the default locale, so on hosts whose locale uses non-Latin digits (ar/fa/bn) the TIME literal comes out as non-ASCII numerals that writeAscii then truncates to garbage. Format with Locale.ROOT, matching the locale-independent encoding the other codecs already use.


rusher:

PR #218, opened 2026-05-26
call inflater.end() in finally in CompressInputStream

jmestwa-coder:
retrieveBuffer creates a new Inflater but only calls end() on the success path, so a server sending a malformed compressed packet leaks the native zlib stream once inflate throws DataFormatException, and the length-mismatch IOException branch leaks the same way. Move end() into a finally so it runs on every path, matching the success-path call already there.


rusher:

PR #219, opened 2026-05-29
use Locale.ROOT in date/time codec text formatting

jmestwa-coder:
DateCodec, TimeCodec and UtilDateCodec build the text-protocol literal with new SimpleDateFormat(pattern) under the default locale, so on hosts whose locale uses non-Latin digits (ar/fa/bn) the numeric fields come out as non-ASCII numerals that writeAscii then truncates to garbage. Format with Locale.ROOT, matching the locale-independent encoding the other codecs already use.


rusher:

PR #220, opened 2026-05-30
encode caching_sha2 clear password as UTF-8

jmestwa-coder:
The full-authentication clear-password path in CachingSha2PasswordPlugin.process (line 225) builds the password bytes with authenticationData.getBytes(), using the platform default charset, while the fast-auth scramble at line 59 and every other password in this package use getBytes(UTF_8). On a JVM whose default charset is not UTF-8 a non-ASCII password is sent with the wrong bytes, so caching_sha2 full auth fails where fast auth would have passed.


rusher:
thanks merged into develop branch, this will be release in next version

PR #221, opened 2026-06-01
synchronize PrepareCache.reset()

jmestwa-coder:
Found this stress-testing the prepare cache during failover. get() and put() are synchronized, but reset() iterates values() and clears the LinkedHashMap with no lock. A reconnect calling resetPrepareCache() while another thread runs put() then races on a non-thread-safe map, giving a ConcurrentModificationException or a corrupted cache. Lock reset() on the same monitor.


rusher:
Thanks for reporting issue. 
After checking possible problem, the best solution was done in https://github.com/mariadb-corporation/mariadb-connector-j/commit/3eb5bcb176f1fa86c454a064b80269fc40b18423
this permit a better alternative than this current suggestion.

PR #222, opened 2026-06-02
use Locale.ROOT when lowercasing host address parameters

jmestwa-coder:
host token in address=(host=..) is lowercased with the default locale, mangling names containing 'I' under tr_TR; pin Locale.ROOT as HostnameVerifier already does.


rusher:
thanks for the report !
it's now merged in develop branch, and will be release in next version (3.5.9)

PR #223, opened 2026-06-03
create gssapi jaas config temp file with owner-only permissions

jmestwa-coder:
The jaas config written in StandardGssapiAuthentication.authenticate is created with File.createTempFile, which leaves it world-readable (rw-r--r-- under the usual umask) in the shared temp dir while its path gets published JVM-wide as java.security.auth.login.config. Switch to Files.createTempFile with rw------- so only the connecting user can read it, falling back to the default on non-POSIX filesystems.


rusher:
thanks for the report and correction. 
this has been merged in maintenance/2.7, maintenance/3.3, maintenance/3.4 and develop branches. 
will be in corresponding releases

PR #224, opened 2026-06-05
use constant-time comparison in validateFingerPrint

jmestwa-coder:
non-constant-time check on the server certificate fingerprint token in validateFingerPrint:
- hashHex is derived from the user's hashed password, the handshake seed and the cert fingerprint, then compared to the server value with String.equals
- String.equals returns on the first differing char, so the time before the connection proceeds or aborts leaks the matching prefix length of a secret-derived token
- a man-in-the-middle fixes the seed and its own fingerprint across reconnects, recovers the token byte by byte and replays it to pass the anti-MitM check without the password
swapped the compare to MessageDigest.isEqual.


rusher:
thanks for the report and correction. This is merged to develop and maintenance/3.4 branches and will be released in next version

PR #225, opened 2026-06-09
match local infile filename case-sensitively

jmestwa-coder:
local infile validation matched the server-requested filename case-insensitively, so a rogue or mitm server can answer a legitimate LOAD DATA LOCAL INFILE with a path that only differs in case and have the client read a different file on case-sensitive filesystems:
- direct match compiled the filename into the regex under Pattern.CASE_INSENSITIVE
- parameterized match compared with equalsIgnoreCase
- statement keywords still need case-insensitive matching, the path does not
wrapped the path in (?-i:) and switched the parameter comparison to equals so the keywords stay case-insensitive but the filename matches exactly.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/message/ClientMessage.java
+++ b/src/main/java/org/mariadb/jdbc/message/ClientMessage.java
@@ -59,9 +59,9 @@ public interface ClientMessage {
             + "\\n"
             + "|\\r"
             + "|\\n"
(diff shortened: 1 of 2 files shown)


rusher:
thanks for reporting [jmestwa-coder](https://github.com/jmestwa-coder), this is merged in all supported version (develop and maintenance/xx branches)

PR #226, opened 2026-06-09
read parsec auth iteration count as unsigned byte

jmestwa-coder:
**Signed read of the parsec iteration factor**
The PBKDF2 iteration factor in the ext-salt is read with `readByte()`, so a server byte >= 0x80 sign-extends past the `iterations > 20` check and feeds a negative shift into `1024 << iterations`; 0xFF lands on a zero iteration count and `PBEKeySpec` then throws an uncaught `IllegalArgumentException`. Reading it with `readUnsignedByte()` keeps the bound effective across the full 0-255 range.


jmestwa-coder:
rebased onto latest main. the earlier red was PooledConnectionTest.testPoolFailover, a flaky pool test unrelated to this change, and main's recent stability fix for it is now in the branch. the parsec patch itself is the same one-line readUnsignedByte change.

rusher:
thanks @jmestwa-coder , this is merged in develop branch, and will be in  3.5.10 version

PR #227, opened 2026-06-13
skip unknown session-track entries from OkPacket session buffer

jmestwa-coder:
Unknown session-track types desync Ok_Packet session-state parsing:
- the default branch skips the entry on the outer packet buffer, not the current sessionStateBuf the other cases use
- a server sending a tracked type the driver does not handle (state change, gtids, transaction info) then drops or misreads a following entry in the same packet, e.g. a SESSION_TRACK_SCHEMA database change
- parse and parseWithInfo both affected

Skip on sessionStateBuf, matching the ColumnDecoder extended-info loop.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/message/server/OkPacket.java
+++ b/src/main/java/org/mariadb/jdbc/message/server/OkPacket.java
@@ -131,7 +131,7 @@ public class OkPacket implements Completion {
                 break;
 
               default:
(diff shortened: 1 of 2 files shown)


jmestwa-coder:
any update?

rusher:
thanks @jmestwa-coder , this is merged in develop branch, and will be in 3.5.10 version

PR #228, opened 2026-06-13
fix unsafe escaping in enquoteLiteral and enquoteNCharLiteral

jmestwa-coder:
**SQL injection in enquoteLiteral / enquoteNCharLiteral**

enquoteLiteral backslash-escapes the embedded quote and never doubles backslashes (its replacement map is passed to Matcher.appendReplacement, which collapses the doubled backslash back to one), so the literal can be closed early: a trailing backslash escapes the closing quote under the default sql_mode, and the `\'` emitted for a quote is a literal backslash plus a terminating quote when sql_mode=NO_BACKSLASH_ESCAPES. enquoteNCharLiteral doubles the quote but leaves backslashes unescaped, breaking out the same way under the default mode, so both now double the backslash and the quote, which is safe under both sql modes and is the doubling the JDBC contract specifies.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/Driver.java
+++ b/src/main/java/org/mariadb/jdbc/Driver.java
@@ -13,7 +13,6 @@ import java.sql.DriverPropertyInfo;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.util.*;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.mariadb.jdbc.client.Client;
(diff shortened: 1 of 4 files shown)


jmestwa-coder:
any update?

rusher:
thanks for report [jmestwa-coder](https://github.com/jmestwa-coder), you're full time on finding issues !
i've created https://jira.mariadb.org/browse/CONJ-1326 for this. 
merged you commit on develop branch + https://github.com/mariadb-corporation/mariadb-connector-j/commit/aa8cb0a87f352eed59a85c48b76b01f63bd53083 commit, that is similar in terms of performance, but a bit faster when escape is needed

PR #229, opened 2026-06-16
mask password in redirect url before logging

jmestwa-coder:
redirect() parses user:password from a server-sent redirect_url and uses them as connection credentials, but logs the raw url:
- logger.info on a successful redirect prints the full url including the password
- logger.error on parse failure and on redirect failure do the same
- a redirect_url like mariadb://user:pwd@host/db therefore writes the cleartext password to the logs
Masked the password with *** at the three log sites.


jmestwa-coder:
any update?

rusher:
thanks @jmestwa-coder jmestwa-coder

it's merged on develop branch, and will be in next release

PR #230, opened 2026-06-16
ignore certificate CN when subjectAltName is present

jmestwa-coder:
HostnameVerifier.verify falls back to the certificate CN even when the cert presents subjectAltName entries of the host's type:
- per RFC 6125 §6.4.4 and the OpenSSL/LibreSSL/BoringSSL default, the CN is only consulted when the certificate carries no SAN entry of the type matching the host (dNSName for DNS hosts, iPAddress for IP hosts)
- a cert with CN=*.mariadb.org and SAN=[DNS:other.org,DNS:www.other.org] is currently accepted for a.mariadb.org, a host its SAN does not cover
- gated the CN fallback on the absence of a matching-type SAN entry, so certs with only iPAddress entries still match DNS hosts on the CN and vice versa

Updated the verifier tests that encoded the old behavior and added coverage for an IP-only SAN cert (CONJ-1327).

### DIFF
--- a/src/main/java/org/mariadb/jdbc/client/tls/HostnameVerifier.java
+++ b/src/main/java/org/mariadb/jdbc/client/tls/HostnameVerifier.java
@@ -279,26 +279,22 @@ public class HostnameVerifier {
       }
 
       // ***********************************************************
-      // RFC 2818 : legacy fallback using CN (recommendation is using alt-names)
+      // RFC 2818 / RFC 6125 : the Common Name is only a legacy fallback and must be
+      // ignored once the certificate carries subjectAltName entries. Reaching here with
+      // a non-empty SAN set means none of them matched, so the host is not authorized.
       // ***********************************************************
+      if (!subjectAltNames.isEmpty()) {
+        throw new SSLException(
+            normalizedHostMsg(lowerCaseHost) + " doesn't correspond to " + subjectAltNames);
+      }
+
       X500Principal subjectPrincipal = cert.getSubjectX500Principal();
       String cn = extractCommonName(subjectPrincipal.getName(X500Principal.RFC2253));
-
       if (cn == null) {
-        if (subjectAltNames.isEmpty()) {
-          throw new SSLException(
-              "CN not found in certificate principal \""
-                  + subjectPrincipal
-                  + "\" and certificate doesn't contain SAN");
-        } else {
-          throw new SSLException(
-              "CN not found in certificate principal \""
-                  + subjectPrincipal
-                  + "\" and "
-                  + normalizedHostMsg(lowerCaseHost)
-                  + " doesn't correspond to "
-                  + subjectAltNames);
-        }
+        throw new SSLException(
+            "CN not found in certificate principal \""
+                + subjectPrincipal
+                + "\" and certificate doesn't contain SAN");
       }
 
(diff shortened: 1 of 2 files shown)


rusher:
Thanks for PR. 

I've created https://jira.mariadb.org/browse/CONJ-1327 for that. rfc 9525 now indicate to not rely on CN, while previous RFC was marking CN use as deprecated. 
since we did have some issues with that previously like https://jira.mariadb.org/browse/CONJ-511, this will be corrected in next 3.6 version, and not as a fix release.

rusher:
openssl default behavior from https://docs.openssl.org/4.0/man3/X509_check_host/

> the default is to ignore the subject DN when at least one corresponding subject alternative names is present.

and behavior can be changed according to X509_CHECK_FLAG_NEVER_CHECK_SUBJECT / X509_CHECK_FLAG_ALWAYS_CHECK_SUBJECT flags

not sure we has to be harsh than openssl

rusher:
even if that is in the standard RFC, 

* OpenSSL's [X509_check_host|https://docs.openssl.org/4.0/man3/X509_check_host/]() still performs a fallback to the Subject CN when no matching SAN is present. The stricter RFC 9525 behavior is available only by explicitly enabling X509_CHECK_FLAG_NEVER_CHECK_SUBJECT. it is not the default.
* LibreSSL documents the same behavior, checking SAN first and falling back to the Subject CN when appropriate. (https://man.uex.se/3/libressl-X509_check_host)
* BoringSSL also documents that its current default still checks the Subject CN for compatibility, although its own header comments describe this as deprecated and recommend callers pass X509_CHECK_FLAG_NEVER_CHECK_SUBJECT to obtain the standards-compliant behavior. https://boringssl.googlesource.com/boringssl/%2B/HEAD/include/openssl/x509.h#3343

so better to follow standard TLS libraries for now, suspending implementation

jmestwa-coder:
makes sense, aligning with the openssl/libressl/boringssl default is fair.

one clarification on that default though: openssl only falls back to the CN when the cert has no SAN of the relevant type. once a DNS SAN is present it ignores the CN even if none of them match, which is the "ignore the subject DN when at least one corresponding SAN is present" line from your first link. so this PR is basically the openssl default already, the only difference is the guard triggers on any SAN rather than a SAN of the matching type (e.g. only IP SANs while checking a DNS host).

i can narrow it to gate on matching-type SAN so it lines up exactly with openssl, or just leave it for CONJ-1327 in 3.6. whatever you prefer.

rusher:
yes, that would be the best , like OpenSSL/LibreSSL/BoringSSL default, and RFC 6125 §6.4.4, no "SAN of matching type", not "no SAN of any type"

i've changed https://jira.mariadb.org/browse/CONJ-1327 description following that

jmestwa-coder:
done, narrowed the guard to the host's san type:

- dns hosts skip the cn only when a dNSName entry is present, ip hosts only when an iPAddress entry is present
- restored the combined cn+san error messages on the fallback paths since they're reachable again
- added a test with an ip-only san cert: dns hosts still match the cn, ip hosts ignore it

also updated the pr description to match the CONJ-1327 wording.

rusher:
merged into develop branch, will be part of next 3.5.10 release.
thanks !

PR #231, opened 2026-06-16
match restrictedAuth plugin list exactly

jmestwa-coder:
restrictedAuth is meant to limit which auth plugins the server can pick, but AuthenticationPluginLoader matches each list entry with type.contains() rather than equality:
- an empty element from a leading or doubled comma makes contains("") true for every plugin, so the restriction is silently disabled
- a partial entry like "mysql" or "password" matches plugins that were never listed
- a server can then request mysql_clear_password and receive the account password in clear text

Switched the match to exact equality.


jmestwa-coder:
any update?

rusher:
thanks for PR @jmestwa-coder 

merged in develop and 3.3, 3.4 maintenance branches.

rusher:
for the record  https://jira.mariadb.org/browse/CONJ-1328

PR #232, opened 2026-06-17
refuse local infile request when parameter value unavailable

jmestwa-coder:
validateLocalFileName accepts any server-requested file when the bound parameter can't be rendered to a literal:
- in the parameterized branch, a null bestEffortStringValue falls through to `return true`, authorizing whatever file the server asked for
- bestEffortStringValue returns null when the value can't be encoded (its contract returns null for streaming parameters), so a malicious or proxy server can make the client read and send an arbitrary local file
- fail closed instead, so a request that can't be matched against the query is refused

regression test added alongside the existing case-sensitivity one.


jmestwa-coder:
any update?

rusher:
thanks for PR. 

https://jira.mariadb.org/browse/CONJ-1329 created for that, and merged in supported branch (develop/3.4/3.3)

PR #233, opened 2026-07-03
validate trustStore chain instead of deferring identity

jmestwa-coder:
An explicit trustStore is put on the lenient deferred-identity path:
- buildTrust wraps the trustStore branch with deferIdentity=true; the equivalent serverSslCert branch uses false
- a server certificate that does not chain to the pinned trustStore is then accepted at the handshake, and under VERIFY_FULL the hostname check is skipped since certFingerprint is set
- both serverSslCert and trustStore are user-provided trust anchors and should validate identically; only the no-explicit-material system fallback is meant to defer

Set the trustStore branch to strict validation, matching serverSslCert.


rusher:
i've created https://jira.mariadb.org/browse/CONJ-1331 for that. 
The PR is merged in develop branch, and will be part of next release

jmestwa-coder:
thanks for taking this into develop, and for filing CONJ-1331

PR #234, opened 2026-07-07
mask sensitive nonMappedOptions when serializing connection url

jmestwa-coder:
the connection url serialization redacts the declared password fields but leaves credentials kept in nonMappedOptions in clear text:
- getUrl() / DatabaseMetaData.getURL() (buildUrl, appendPropertiesParameter) print nonMapped options verbatim, so AWS IAM secretKey and PAM password2/password3 leak while password already shows ***
- toConf()'s "Unknown options" block (appendUnknownOptions) has the same gap
these urls are commonly logged by pools, orms and monitoring. mask nonMapped options whose name contains password or secret, matching the existing SECURE_FIELDS redaction.

### DIFF
--- a/src/main/java/org/mariadb/jdbc/Configuration.java
+++ b/src/main/java/org/mariadb/jdbc/Configuration.java
@@ -1047,7 +1047,9 @@ public class Configuration {
             entry ->
                 new AbstractMap.SimpleEntry<>(
                     entry.getKey().toString(),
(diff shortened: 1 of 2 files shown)


rusher:
merged in develop branch with correction : DatabaseMetaData.getURL() is expect to return exact value with passwords. All pools JNDI implementation use setURL/getURL, that cannot be changed

PR #235, opened 2026-07-08
cap length-encoded field length before narrowing to int in row decoders

jmestwa-coder:
Two row decoders narrow a length-encoded field length to int with an unchecked cast:
- TextRowDecoder.setPosition keeps only the low 32 bits, so a server length ending 0xFFFFFFFF becomes -1, which is NULL_LENGTH, and a non-null column reads back as SQL NULL
- BinaryRowDecoder.setPosition has the same cast on the 8-byte form; other high-bit lengths become a negative int that reaches new byte[] in the codecs
Read the full value and reject anything outside 0..Integer.MAX_VALUE.


rusher:
merged in develop branch and applied on supported branch. 
thanks @jmestwa-coder

PR #236, opened 2026-07-10
escape identifiers in updatable result set generated sql

jmestwa-coder:
**Identifiers not escaped in updatable result set queries**
The insert/update/delete/select builders in `UpdatableResult` wrap the schema, table and column names in backticks without doubling backticks inside the name, so a name containing a backtick (allowed by MariaDB, and reachable through a result-set column alias) closes the identifier early and injects SQL into the generated statement. Routed those identifiers through a helper that doubles backticks, matching `enquoteIdentifier` and the savepoint / `CREATE DATABASE` paths; behavior for backtick-free names is unchanged.


rusher:
filled as [CONJ-1340](https://jira.mariadb.org/browse/CONJ-1340)
merged in develop branch and cherry picked to maintained branches

PR #237, opened 2026-07-20
propagate Connection.setReadOnly() to server via opt-in readOnlyPropagatesToServer

mariadb-VasilyKozhukhovskiy:
Connection.setReadOnly(true) was a no-op on the server for single-host connections and only affected primary/replica routing on multi-host ones. The server session was never told, so an engine that can use that the session has read-only state could not take its optimistic read-only transaction path.

Add a readOnlyPropagatesToServer connection option (default false). When enabled, setReadOnly() additionally issues SET SESSION TRANSACTION READ ONLY / READ WRITE to the server on every topology (single-host and multi-host), on servers >= 5.6.5. When disabled, behavior is unchanged: no statement is sent and only the existing routing applies.

- StandardClient: send the statement when the option is enabled.
- MultiPrimaryClient: delegate to the current client and replay the target direction on failover via syncReadOnlyState().
- MultiPrimaryReplicaClient: set the requested direction before syncNewState() so replay uses the correct flag value.

Known limitation: in a replica topology with no reachable replica, the connection stays on the primary and the primary is not put in read-only mode.

Tests cover all three clients with the option on and off.


rusher:
closing, done with https://github.com/mariadb-corporation/mariadb-connector-j/commit/cd0b6fa162ffeb53c0925c47bafd4621eff5e66f. Not using PR because multi doesn't need that : replica connections are set to read only, and setReadOnly then only switch to write or read connection +  avoid executing command if not needed

PR #240, opened 2026-09-17
Datasources no longer register the driver in DriverManager (2.x behaviour)

mario-schwede-hivemq:
Hello, and thanks for maintaining the connector.

`org.mariadb.jdbc.Driver` registers itself in `DriverManager` from its static initializer, and `DriverManager` keeps that instance for the lifetime of the JVM — which pins the class loader that loaded it. That is the usual leak when an application server redeploys an application bundling the connector.

In 2.x this never affected datasource users: `MariaDbDataSource.getConnection()` went straight to `MariaDbConnection.newConnection(...)`, and `Pool` never referenced `Driver` at all. Since 3.0 both paths go through `Driver.connect(...)`, so simply using a `MariaDbDataSource` is enough to register the driver. This PR restores the earlier behaviour.

The implementation moves into a new public `NonRegisteringDriver`, and `Driver` keeps only the static initializer that registers it. This mirrors MySQL Connector/J, where `com.mysql.cj.jdbc.Driver extends NonRegisteringDriver` carries nothing but the registration and `MysqlDataSource` connects through `NonRegisteringDriver.connect(...)` — so the name and the split should already be familiar. `Driver` remains the `java.sql.Driver` service provider, so `Class.forName` and the SPI are unchanged, and the static helpers (`enquoteLiteral` and friends) still compile for existing callers.

Both paths are covered by `doesNotRegisterDriver` tests, which load the connector in a class loader that refuses to define `org.mariadb.jdbc.Driver` and then open a real connection. The full suite passes locally.

Would you like a CONJ ticket for this? Happy to open one if you prefer it tracked there, and happy to adjust anything in the change itself.


rusher:
thanks for the detailed explaination and the PR. 
I've created [CONJ-1355](https://jira.mariadb.org/browse/CONJ-1355) to document that, and merged this PR into develop branch. This will be in next version

mario-schwede-hivemq:
Thanks for merging this so quickly, and for opening CONJ-1355.

One follow-up: On 2.7.15, we call `SchedulerServiceProviderHolder.close()` when we shut down. That stops the driver's threads and frees the class loader. 3.x has nothing like it.

`org.mariadb.jdbc.client.util.SchedulerProvider` keeps the timeout scheduler in a `static` field, with one core thread that never times out, and it has no `close()` or `shutdown()`. So once the `MariaDb-timeout` thread starts, nothing can stop it, and the class loader stays alive. We talk to MySQL servers, so `canUseServerTimeout` is always false and the first query with a timeout starts that thread.

I don't want to suggest a fix. Adding a `close()` would give us back what 2.x had, but a global static field is not great either way, and you may prefer to tie the scheduler to the connection or the pool instead. One thing that is worth knowing: bringing back the 2.x `DriverAction` would not be enough on its own, because after this PR datasource users never register, so it would never run.

Happy to open an issue with the details if that helps.

rusher:
thanks for reporting that as well. https://jira.mariadb.org/browse/CONJ-1358 has been created for that, and handle with this commit : https://github.com/mariadb-corporation/mariadb-connector-j/commit/5d20f7f1537e01373ceb72f6b87f206b4a9cb39d

this will be in 3.5.11 version.  

this is available in snapshot version :
```
<repositories>
  <repository>
    <id>central-snapshots</id>
    <url>https://central.sonatype.com/repository/maven-snapshots/</url>
    <releases><enabled>false</enabled></releases>
    <snapshots>
      <enabled>true</enabled>
      <updatePolicy>always</updatePolicy>
    </snapshots>
  </repository>
</repositories>

<dependency>
  <version>3.5.11-SNAPSHOT</version>
</dependency>
```

